PR #32, opened 2019-11-07
Initial pass over implementing Activation Functions.

maximkrouk:
Just a few functions for now.

### DIFF
--- a/Sources/Real/ActivationFunctions.swift
+++ b/Sources/Real/ActivationFunctions.swift
@@ -1,0 +1,31 @@
+//===--- ActivationFunctions.swift ----------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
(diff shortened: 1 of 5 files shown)


xwu:
So, the category of "activation functions" overlaps with that of elementary functions. I'm not sure of the utility of defining ReLU for a general-purpose numerics library. As for `sigmoid`, what you've got here is the standard logistic function, which I think would be fantastic as an extension to `Real`.

stephentyrone:
Echoing what @xwu said, these probably go in a separate module rather than in `Real` (which is fine!)

A bigger question is that I'm not sure how useful it is to define these for *scalar* values rather than contiguous buffers, since the overwhelming use case in ML is on buffers of data (and you often want to fold them into an adjacent operation, but let's ignore that for now).

maximkrouk:
Well, so should I change anything or maybe close this pr?)

PR #57, opened 2019-11-08
Added LinuxMain.swift so numerics can be used on Linux

tachoknight:
Wanting to use this library on Linux, running `swift run --repl` fails because `LinuxMain.swift` is not  present under the Tests directory. Adding this file allows the library to work on Linux.

### DIFF
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -1,0 +1,25 @@
+import XCTest
+
+@testable import ComplexTests
+@testable import RealTests
+
+XCTMain([
+    testCase(ArithmeticBenchmarkTests.testDivisionByConstant),
+    testCase(ArithmeticBenchmarkTests.testReciprocal),
+    testCase(ArithmeticBenchmarkTests.testDivisionByConstantC),
+    testCase(ArithmeticBenchmarkTests.testMultiplication),
+    testCase(ArithmeticBenchmarkTests.testMultiplicationC),
+    testCase(ArithmeticBenchmarkTests.testDivision),
+    testCase(ArithmeticBenchmarkTests.testDivisionC),
+    testCase(ArithmeticBenchmarkTests.testDivisionPoorScaling),
+    testCase(ArithmeticBenchmarkTests.testDivisionPoorScalingC),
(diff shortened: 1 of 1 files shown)


stephentyrone:
`swift test --enable-test-discovery`

tachoknight:
Hm, when I try that I get `**error: no tests found; create a target in the 'Tests' directory**` 
Creating an empty `LinuxMain.swift` file prevents the error, but I figured to actually include the individual tests for full coverage.

stephentyrone:
> Hm, when I try that I get **error: no tests found; create a target in the 'Tests' directory**

@aciidb0mb3r is that expected?

SusanDoggie:
exec the following on Mac 

swift test --generate-linuxmain

ole:
It looks like `swift build` fails on Linux when `LinuxMain.swift` is missing unless you specify `swift build --enable-test-discovery`. See this forum thread: https://forums.swift.org/t/make-test-discovery-on-by-default/30321.

The contents of `LinuxMain.swift` are not important. It can be empty or contain a `fatalError` pointing people to `--enable-test-discovery`, but the file must exist for now.

stephentyrone:
Hi all--as a follow-on, I've replaced this LinuxMain.swift with a file containing only a `fatalError( )` with a message directing users to `--enable-test-discovery`. Can someone please confirm that this works as intended on Linux, and then I'll cut a 0.0.1 tag?

ole:
@stephentyrone I tried running the tests from macOS in a Ubuntu 18.04/Swift 5.1 Docker container. The LinuxMain thing worked for fixing `swift build`, but there was another issue when building the tests. I sent PR #74 for that.

When #74 is merged, the tests run correctly.

By the way, if you install [Docker for Mac](https://docs.docker.com/docker-for-mac/install/), you can run open a Linux terminal with Swift installed by running this command:

```sh
docker run --rm --privileged --interactive --tty --volume $(pwd):/src --workdir "/src" swift:latest
```

This will map your current working directory to `/src` in the Docker container. Then you can use `swift build` and `swift test` to test things on Linux (you may have to make a clean build because SwiftPM can get confused by having build artifacts from both macOS and Linux).

Issue #62, opened 2019-11-10
Slight tweaks to Complex


CTMacUser:
I just took a look at this project, and have some suggestions for slight tweaks.

- `magnitude` should return `RealType.magnitude` instead of `RealType`.
- `IntegerLiteralType` should alias to `RealType`’s version.
- The `Codable` conditional conformance could be split into separate `Encodable` and `Decodable` conditional conformances.
- The `debugDescription` could be unconditional.

        extension Complex: CustomDebugConvertible {
            public var debugDescription: String {
                String(describing: Self.self) + “(“ + String(reflecting: x) + “, “ + String(reflecting: y) + “)”
            }
        }

SusanDoggie:
It seems that RealType.Magnitude always equal to RealType

https://github.com/apple/swift/blob/f5f214de21b24627b9499a1581a64052f1235f09/stdlib/public/core/FloatingPoint.swift#L162

stephentyrone:
> I just took a look at this project, and have some suggestions for slight tweaks.
> 
> * `magnitude` should return `RealType.magnitude` instead of `RealType`.

As Susan noted, these types are always the same, so this is only a spelling change.

> * `IntegerLiteralType` should alias to `RealType`’s version.

I don't think that this actually changes anything with the way integer literals are currently implemented in the compiler, but I don't think that it hurts anything either.

> * The `Codable` conditional conformance could be split into separate `Encodable` and `Decodable` conditional conformances.

I'm not sure if this really gains us anything or not (I expect a Real type that only conforms to one of the two to be quite unusual), but it also doesn't cost anything, so sure.

> * The `debugDescription` could be unconditional.
>   ```
>     extension Complex: CustomDebugConvertible {
>         public var debugDescription: String {
>             String(describing: Self.self) + “(“ + String(reflecting: x) + “, “ + String(reflecting: y) + “)”
>         }
>     }
>   ```

Sure.

stephentyrone:
@CTMacUser I think that all of the points raised here are resolved, OK to close this?

CTMacUser:
You're not planning to update `IntegerLiteralType`?

My sample had "`String(describing: Self.self)`" replacing "`Complex<\(RealType.self)>`" because I found out that when Swift prints out a type, and the type is generic, it'll include the generic arguments along with the base name.

stephentyrone:
> You're not planning to update IntegerLiteralType?

This change has been made:
[...]

PR #65, opened 2019-11-10
Split Codable conformance of Complex into Encodable and Decodable. NFC.

stephentyrone:
Also make conformance to CustomDebugStringConvertible unconditional using String(reflecting:) on x and y.


stephentyrone:
Tweaks suggested in https://github.com/apple/swift-numerics/issues/62

PR #69, opened 2019-11-11
Defer to libm cbrt for root(x,3).

stephentyrone:
This lets us get exact cases right only platforms where cbrt(x) has sub-ulp accuracy. It doesn't get all the other non-power-of-two roots quite right, and is still dependent on the host system, but 2 and 3 are the most common roots people take by far, so it's a good first step.

### DIFF
--- a/Sources/Real/ScalarConformances.swift
+++ b/Sources/Real/ScalarConformances.swift
@@ -204,7 +204,7 @@ extension Float80: Real {
     // where we can fallback on cbrt.
     if n == 3 { return libm_cbrtl(x) }
     // TODO: this implementation is not quite correct, because either n or
-    // 1/n may be not be representable as Double.
+    // 1/n may be not be representable as Float80.
     return Float80(signOf: x, magnitudeOf: libm_powl(x.magnitude, 1/Float80(n)))
   }
   
--- a/Sources/Real/ScalarConformances.swift
+++ b/Sources/Real/ScalarConformances.swift
@@ -63,7 +63,7 @@ extension Float: Real {
     // where we can fallback on cbrt.
     if n == 3 { return libm_cbrtf(x) }
     // TODO: this implementation is not quite correct, because either n or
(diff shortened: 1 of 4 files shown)


stephentyrone:
CC @Roman-Kerimov

Sources/Real/ScalarConformances.swift
```diff
@@ -57,6 +57,16 @@ extension Float: Real {
     return libm_powf(x, Float(n))
   }
   
+  @_transparent public static func root(_ x: Float, _ n: Int) -> Float {
+    guard x >= 0 || n % 2 != 0 else { return .nan }
+    // Workaround the issue mentioned below for the specific case of n = 3
+    // where we can fallback on cbrt.
+    if n == 3 { return libm_cbrtf(x) }
+    // TODO: this implementation is not quite correct, because either n or
+    // 1/n may be not be representable as Double.
```
xwu:
```suggestion
    // 1/n may be not be representable as Float.
```

Sources/Real/ScalarConformances.swift
```diff
@@ -178,6 +198,16 @@ extension Float80: Real {
     return libm_powl(x, Float80(n))
   }
   
+  @_transparent public static func root(_ x: Float80, _ n: Int) -> Float80 {
+    guard x >= 0 || n % 2 != 0 else { return .nan }
+    // Workaround the issue mentioned below for the specific case of n = 3
+    // where we can fallback on cbrt.
+    if n == 3 { return libm_cbrtl(x) }
+    // TODO: this implementation is not quite correct, because either n or
+    // 1/n may be not be representable as Double.
```
xwu:
```suggestion
    // 1/n may be not be representable as Float80.
```

stephentyrone:
@Roman-Kerimov the clang `__builtin` namespace is partially documented here: https://clang.llvm.org/docs/LanguageExtensions.html#builtin-functions

PR #72, opened 2019-11-12
Remove broken link in Real module readme

ole:
The link target for `[Real]` wasn't defined. I decided toremove the link altogether because the readme would essentially link to itself.

Let me know if you'd rather keep the link (presumably it should link to <https://github.com/apple/swift-numerics/tree/master/Sources/Real>?).


stephentyrone:
Thanks @ole !

Issue #51, opened 2019-11-08
Iron out semantics of Complex(length:phase:)


NevinBR:
https://github.com/apple/swift-numerics/blob/5d4368257efebefff658d7c0fa06297abd040d5d/Sources/Complex/Complex.swift#L486

Currently, the polar initializer `Complex.init?(length:phase:)` is failable. However, `Complex` already has a representation for invalid results, namely `(.nan, .nan)`, which is treated as the point at infinity.

Thus, it would be reasonable for the polar initializer to be non-failable, and always produce `(length * .cos(phase), length * .sin(phase))`. In the case that either `length` or `phase` is non-finite, the result would be `Complex.infinity`.

This is the same result one would get by first calculating the sine and cosine of the phase as a `RealType` then constructing the `Complex` result from its components, so it seems reasonable that it’s what users will expect.

stephentyrone:
`.nan, .nan` is not a representation for invalid results; it's treated as another infinity.

This is somewhat surprising. The advantage of this choice is that it lets us use the obvious expression for multiplication, (rr-ii, ri+ri) rather than having to detect infinities and handle them specially (if you try to distinguish infinity and nan, then you have to deal with the fact that (inf,0) * (0, i) would give you (nan, inf), which ... is that a nan or an infinity? This is a mess for C and C++,it results in everyone who cares about performance using their own multiplication instead of the language-provided one).

There are also downsides to this approach, but it pushes some of the weirdness out of basic arithmetic and onto less-frequently-used operations. I'm open to changing it, but there's a definite tension.

NevinBR:
> `.nan, .nan` is not a representation for invalid results; it's treated as another infinity.

I understand that it is treated as an infinity, yet I maintain that it *is* a representation of invalid results. Or at least, `Complex.infinity` in general is currently used to represent invalid results:

`.infinity - .infinity` is undefined, but represented by `.infinity`
`.infinity / .infinity` is undefined, but represented by `.infinity`
`(0, 0) / (0, 0)` is undefined, but represented by…wait, that’s `(0, 0)`? Is this intentional? I would expect `.infinity`.

stephentyrone:
> > `.nan, .nan` is not a representation for invalid results; it's treated as another infinity.
> 
> I understand that it is treated as an infinity, yet I maintain that it _is_ a representation of invalid results. Or at least, `Complex.infinity` in general is currently used to represent invalid results:
> 
[...]

PR #77, opened 2019-11-20
Address behavior of Complex.init(length:phase:)

stephentyrone:
Instead of being failable, adopt the following policy:

- If length is zero or infinity, the phase does not matter.
- Otherwise, the phase must be finite or a precondition failure occurs.

Note that this is a source-breaking change.

Discussed here: https://github.com/apple/swift-numerics/issues/51

### DIFF
--- a/Sources/Complex/Complex.swift
+++ b/Sources/Complex/Complex.swift
@@ -472,7 +472,7 @@ extension Complex {
     (length, phase)
   }
   
-  /// A complex value specified with  polar coordinates.
+  /// Creates a complex value specified with polar coordinates.
   ///
   /// Edge cases:
   /// -
--- a/Sources/Complex/Complex.swift
+++ b/Sources/Complex/Complex.swift
@@ -472,26 +472,39 @@ extension Complex {
     (length, phase)
   }
   
-  /// Constructs a complex value from polar coordinates.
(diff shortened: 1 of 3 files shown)


stephentyrone:
CC @NevinBR, @xwu, @susandoggie

Sources/Complex/Complex.swift
```diff
@@ -472,26 +472,39 @@ extension Complex {
     (length, phase)
   }
   
-  /// Constructs a complex value from polar coordinates.
+  /// A complex value specified with  polar coordinates.
```
rxwei: I think initializer documentation usually starts with `Creates a/an`.
```suggestion
  /// Creates a complex value specified with polar coordinates.
```

NevinBR:
Looks good except it should accept `-.infinity` for the length.

stephentyrone:
Hrm. We don't *need* it to allow round-tripping, but it also doesn't hurt anything.

NevinBR:
Do we care about trapping on `length == .nan`, or should we just make the condition be:

```
length == 0 || !length.isFinite
```

stephentyrone:
Yes, I deliberately want to trap that. I don't see a good use case for supporting it.

Issue #64, opened 2019-11-10
Usage of "unsafe" does not match Swift’s standards


NevinBR:
In the standard library, Swift uses “unsafe” to denote operations that are memory-unsafe. This sets an expectation which I think we should respect and follow.

Currently, `Complex` has a member named `unsafeLengthSquared`. This operation is memory-safe, so in accordance with Swift’s established precedent it should not have “unsafe” in its name.

I propose renaming it to `lengthSquared`, and documenting that the calculation could overflow (or underflow).

xwu:
We can call it `naiveLengthSquared` or `uncheckedLengthSquared`, but I do also think it’s also fine not to prefix it with anything as the square of a value is pretty self-evidently going to have the possibility of overflowing.

xwu:
I will say, though, that if `magnitude` is going to be used for the infinity norm, we might as well be precise and name these things `euclideanNorm`, `euclideanNormSquared`, etc.

stephentyrone:
There’s not yet any established precedent for naming numerical operations that produce useless results for poorly-scaled inputs, but which are sometimes needed for performance reasons. We’re effectively creating it.

It’s reasonable to argue that we don’t need “unsafe”. I think using “lengthSquared” alone is a mistake—we want a uniform prefix that indicates a fast algorithm that does not correctly handle poorly scaled data, and that prefix should be neither the empty string nor something like “fast”. I’m open to something other than “unsafe”, though.

davedelong:
> ...we want a uniform prefix that indicates a fast algorithm that does not correctly handle poorly scaled data, and that prefix should be neither the empty string nor something like “fast”. I’m open to something other than “unsafe”, though.

Small suggestion: the standard library uses `underestimated` (like `underestimatedCount`) in a handful of places.

stephentyrone:
Underestimated isn’t semantically correct in general, unfortunately.

NevinBR:
Am I correct that `x*x + y*y` will always be within one `ulp` of the closest representable value to the actual length squared?

stephentyrone:
A 1-ulp error bound is not totally obvious. Without thinking very hard because it's Sunday evening, the actual bound might be something like 1.25 ulp, but it's definitely quite small.

Sajjon:
I think we should change it to a throwing function:

```swift
public var unsafeLengthSquared: RealType {
    x*x + y*y
  }
```

It is not constant time performance on this computed property, something that is encouraged for computed properties.

[...]

PR #78, opened 2019-11-20
Rename unsafeLengthSquared -> lengthSquared

stephentyrone:
Addresses https://github.com/apple/swift-numerics/issues/64

Also further tweak polar init for Complex to allow (-infinity, θ) when θ is not finite (follow-on from https://github.com/apple/swift-numerics/pull/77)


stephentyrone:
CC @NevinBR

PR #82, opened 2019-11-24
Removed `guard x >= 0` from `pow(x, y)` implementations

NevinBR:
It is perfectly reasonable to call `pow` with a negative base. If the exponent is an integer the calculation will succeed, and if not then the result of the `libm_pow` call will correctly be `.nan`. We don’t need to prohibit the valid case when we can simply call the library function unconditionally.

### DIFF
--- a/Sources/Real/ScalarConformances.swift
+++ b/Sources/Real/ScalarConformances.swift
@@ -43,11 +43,7 @@ extension Float: Real {
   @_transparent public static func gamma(_ x: Float) -> Float { return libm_tgammaf(x) }
   @_transparent public static func log2(_ x: Float) -> Float { return libm_log2f(x) }
   @_transparent public static func log10(_ x: Float) -> Float { return libm_log10f(x) }
-  
-  @_transparent public static func pow(_ x: Float, _ y: Float) -> Float {
-    guard x >= 0 else { return .nan }
-    return libm_powf(x, y)
-  }
+  @_transparent public static func pow(_ x: Float, _ y: Float) -> Float { return libm_powf(x, y) }
   
   @_transparent public static func pow(_ x: Float, _ n: Int) -> Float {
     // TODO: this implementation is not quite correct, because n may be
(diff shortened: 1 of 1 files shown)


stephentyrone:
`pow(_:Self,_:Self)` binds the analytic function `exp(y * log(x))`; this is undefined for negative x, even if `y` is an integer (these are the semantics of the IEEE 754 `powr` operation). This is a deliberate choice to make the integer y cases be `nan`.

NevinBR:
I don’t understand the rationale.

First, why would we copy the behavior of `powr` rather than `pow`? The latter does allow negative-raised-to-integer.

Second, the branch cut along the negative real axis does not prevent `log(x)` from returning a valid result, it simply means there is a discontinuity crossing the cut. I would fully expect that `Complex.log(-1)` should return `.pi * .i`.

Consequently, `exp(y * log(x))` should be valid for any complex numbers `x` any `y`. If `x` is negative-real and `y` is an integer, then the result will be real. Hence, `pow(x, y)` should be real, and thus return a correct result as an operation on real numbers

stephentyrone:
IEEE 754 `powr` and `pown` functions provide much simpler to document and to explain edge cases; `pow` is a weird legacy case caused by languages not providing both overloads in the past and needing to shoehorn both behaviors into a single operation.

`Complex.log(-1)` does return `.pi * .i.` (or will, when implemented), yes. `Float.log(-1)` returns `.nan`. We're talking about Real types in this PR, right?

NevinBR:
> We're talking about Real types in this PR, right?

Yes. We are also talking about integer exponents.

Raising a base to a positive integer power is defined for anything with a multiplication. Raising to a negative integer power is defined for anything with a reciprocal as well.

If we keep the current behavior where, eg. `Double.pow(-2, -2)` returns `.nan` instead of `0.25`, then I foresee this being a common and recurring source of frustration among users.

stephentyrone:
What is the circumstance, in practice, in which you intend the repeated-multiplication (of reciprocal) operation, but use a floating-point value instead of an integer to represent the exponent? Why wouldn't you write `Double.pow(-2, -2)` (or more likely, `Double.pow(x, -2)` instead?)

PR #83, opened 2019-11-30
Initial pass at pow(Self, Int) for Float and Double

stephentyrone:
More clever implementation is certainly possible (and desired in the long-term), but this basically gets us to the desired semantics without undue performance penalty for the common use cases.

### DIFF
--- a/Sources/Real/ScalarConformances.swift
+++ b/Sources/Real/ScalarConformances.swift
@@ -50,11 +50,21 @@ extension Float: Real {
   }
   
   @_transparent public static func pow(_ x: Float, _ n: Int) -> Float {
-    // TODO: this implementation is not quite correct, because n may be
-    // rounded in conversion to Float. This only effects very extreme cases,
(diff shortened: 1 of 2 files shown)


NevinBR:
Very nice. It might be worth mentioning that the `Float` and `Double` versions will both work properly on hypothetical future platforms where `Int` is more than 64 bits (because all `x != 1` will overflow or underflow before `n` saturates 63 bits), but the `Float80` version will not.

…or it might not be worth mentioning.

stephentyrone:
On a hypothetical future platform where `Int` is more than 64 bits:
- `Float80` doesn't exist, because it's only available on `i386` and `x86_64`.
- swift itself will have a whole bunch of issues.
- the tests will immediately start failing if `Float80` does somehow exist.
So this is mostly theoretical. But I'll add a note anyway =)

stephentyrone:
This change (without the note) is tagged as 0.0.3.

Dindin1992 (commented):
Ok

PR #84, opened 2019-12-17
[Sketch]Arbitrary-precision BigInt

### LINKED ISSUE #5: Arbitrary-precision bigints
Add a flexibly-sized bigint type that grows as required, conforming to `BinaryInteger`.

rothomp3:
Fixes #5 

This is the step 2 "Sketch" PR. It's actually very complete, but obviously perhaps not ideal in style and documentation (since there is none of the latter). Probably needs more tests too, though there are some.

The basic architecture is based on Knuth's algorithms for this purpose. An integer is broken up into "digits" where each digit is in base 64 (or base 32 on 32-bit systems, which are supported). This maps nicely onto the `BinaryInteger` protocol since we can simply use an array of `Uint` for the words. We're also using 2's complement internally, which is super convenient because then for any pair of `BigInt`s that are each only one word long, we can simply delegate to the regular operators for a minimal overhead.


Sources/NumericsShims/include/NumericsShims.h
```diff
@@ -143,6 +143,18 @@ HEADER_SHIM float libm_log10f(float x) {
   return __builtin_log10f(x);
 }
 
+HEADER_SHIM float libm_ceilf(float x) {
```
stephentyrone: These don't appear to be used anywhere?

rothomp3: No, I only use the `double` version in the BigInt implementation, but I thought for the sake of completeness I would add all the variants. If that's wrong, I'd be happy to take 'em out!

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,736 @@
...
+    internal var _isNegative: Bool {
+        return words[words.endIndex - 1] > Int.max
+    }
+    
+    public init<T>(_ source: T) where T : BinaryFloatingPoint {
+        let isNegative = source < 0.0
+        var float = Double(isNegative ? -source : source)
+        var words = Words()
+        while float > 0.0 {
+            let digit = UInt(libm_remainder(float, Double(UInt.max) + 1.0))
```
stephentyrone: This doesn't do what you want it to; `remainder(a,b)` produces a value in `[-b/2, b/2]`, but you need a result in `[0,b)`. This will trap for many inputs when you try to init `UInt` with a negative value. The standard library gives you everything you need here:
```
let radix = T(sign: .plus, exponent: T.Exponent(UInt.bitWidth), significand: 1)
repeat {
  digit = UInt(float.truncatingRemainder(dividingBy: radix))
  words.append(digit)
  float = (float / radix).rounded(.towardZero)
} while float != 0
```
You need to handle a bunch of edge cases that this currently ignores, too: infinity, NaN, and the case where UInt.max > T.greatestFiniteMagnitude (in which case you get a single word that is just `UInt(float)`.)

Sources/NumericsShims/include/NumericsShims.h
stephentyrone: As noted at the use sites, you don't need any of these.

rothomp3:
Ok, new commit to address the floating-point feedback. There doesn't seem to be any easy way to test the one with the `T.greatestFiniteMagnitude` absent a Float16 type in the stdlib.

Sajjon:
@rothomp3 Friendly reminder that you can "bulk" manage "access control" using extensions. Extension also give more clear groupings (especially when used with `// MARK` (which renders nicely in Xcode 11)).

I recommend declaring `BigInt` in one scope with just the stored properties (and if applicable any "designated" initializer. I write "designated" since structs does not have that formally, but sometimes conceptually).

So:
```swift
public extension BigInt {
    func foo() {}

    func bar() {}
}
``` 

Which you can at any point change to `internal` just by changing one statement, as opposed to what you currently have:

```swift 
extension BigInt {
    public func foo() {}

   public func bar() {}
}
```

Sajjon:
@rothomp3 I have not yet looked through your PR in detail, I'm sure it is great! But I just wanted to give you two references - which you surely know about, but if not they might be useful:

[Apple's own BigInt Prototype](https://github.com/apple/swift/blob/master/test/Prototypes/BigInt.swift)

[attaswift/BigInt](https://github.com/attaswift/BigInt)

rothomp3:
> @rothomp3 I have not yet looked through your PR in detail, I'm sure it is great! But I just wanted to give you two references - which you surely know about, but if not they might be useful:
> 
> [Apple's own BigInt Prototype](https://github.com/apple/swift/blob/master/test/Prototypes/BigInt.swift)
> 
> [attaswift/BigInt](https://github.com/attaswift/BigInt)

So both of those use additional storage for the sign, and all the extra math required for that, instead of using 2's complement to store the BigInts the same way the language/CPU already store signed integers of 64 bits or less. I believe my way more effectively leverages the hardware.

xwu:
> I believe my way more effectively leverages the hardware.

Do we have benchmarks that demonstrate this? (The BigInt implementations I’ve seen use sign-magnitude, and I don’t think it’s principally for ease of implementation.)

stephentyrone:
> > I believe my way more effectively leverages the hardware.
> 
> Do we have benchmarks that demonstrate this?

It makes (approximately) no difference. This is pretty easy to see this without benchmarks.

First off, note that converting from one to the other is O(n), and every arithmetic and bitwise operation defined on FixedWidthInteger is at least O(n), so there can be no difference in asymptotic performance--you could always just convert, do the operation, and convert back.

So we're only interested in the the constants.
- bitwise operations are exactly the same either way.
- addition and subtraction are slightly *easier* in the 2s complement representation.
- multiplication requires a tiny bit more bookkeeping (I really mean a tiny bit--the difference is essentially immaterial, a couple additional operations on Int / UInt for schoolbook multiplication, and essentially no additional operations for fast multiplication algorithms).
- division is just multiplication and subtraction, so doesn't matter.

This approach is fine, and most closely matches the semantics of `FixedWidthInteger`, so it's the easiest to use in some ways.

@rothomp3, sorry, I've been busy this week, but I'll give this a more thorough review next week and once we're to a good place I'll set up a branch to merge it onto for further development.

stephentyrone:
> @rothomp3 Friendly reminder that you can "bulk" manage "access control" using extensions.

We deliberately don't do this in the standard library, and I'm going to follow that style with Swift Numerics to facilitate moving code into the standard library if evolution wants some of these features in the future. I know that people have strong opinions about it, but ultimately it doesn't matter very much. Please keep explicit access control for now for simplicity.

Wildchild9:
I was looking at the code for the `BigInt` failable initializer that takes a `String` and I had a couple of thoughts:
- We should allow for the presence of a prefix `+` operator
- We can do remove redundant operations by dropping leading zeros

I think the initializer should be modified to something along the lines of the following:
```swift 
public init?(_ description: String) {
    guard let firstCharacter = description.first else { return nil }
    var description = description
    var isNegative = false
    
    if firstCharacter == "-" {
        guard description.count > 2 else { return nil }
        isNegative = true
        description.removeFirst()
    } else if firstCharacter == "+" {
        guard description.count > 2 else { return nil }
        description.removeFirst()
    }
    
    let validDigits = Set("0123456789")
    guard description.allSatisfy({ validDigits.contains($0) }) else { return nil }
    var result: BigInt = 0
    
    for (i, char) in description.drop(while: { $0 == "0" }).reversed().enumerated() {
        result += BigInt(char.wholeNumberValue!) * pow(10, BigInt(i))
    }
    
    if isNegative {
        result = -result
    }
    
    words = result.words
}
```

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,750 @@
...
+    }
+    
+    @inlinable
+    public func quotientAndRemainder(dividingBy rhs: BigInt) -> (quotient: BigInt, remainder: BigInt) {
+        return BigInt._div(lhs: self, rhs: rhs)
+    }
+    
+    @inlinable
+    public func signum() -> BigInt {
+        return _isNegative ? -1 : 1
```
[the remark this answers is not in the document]
Wildchild9: Yes

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,750 @@
...
+        var description = description
+        var isNegative: Bool = false
+        if description.count > 1 {
+            if firstCharacter == "-" {
+                isNegative = true
+                description.removeFirst()
+            } else if firstCharacter == "+" {
+                description.removeFirst()
+            }
+        }
```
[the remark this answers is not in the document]
rothomp3: So I tried to find a good way to use Substring instead, but it all looked way uglier than this. Hopefully you're not creating BigInts from Strings in a tight inner loop.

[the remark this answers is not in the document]
Wildchild9: @benrimmington, here is an updated version of the initializer. While I would not say that it is simplified, it does have a few advantages of the current version:
1. It does not make an unnecessary copy of the `description` 
2. It removes leading zeros from `description` before the conversion process from `String` to `BigInt` character by character to avoid performing redundant operations.

```swift
public init?(_ description: String) {
    // Ensures `description` is not empty.
    guard let firstCharacter = description.first else { return nil }
    
    var hasPrefixOperator = false
    var isNegative: Bool = false
    
    switch firstCharacter {
    case "-":
        isNegative = true
        fallthrough
    case "+":
        guard description.count > 1 else { return nil }
        hasPrefixOperator = true
    default:
        break
    }
    
    let description = hasPrefixOperator ? description.dropFirst() : description[...]
    
    let digits = Set("0123456789")
    guard description.allSatisfy({ digits.contains($0) }) else { return nil }

    var result: BigInt = 0
    for (i, char) in description.drop(while: { $0 == "0" }).reversed().enumerated() {
        // We're okay to force unwrap here because of the guard above.
        result += BigInt(char.wholeNumberValue!) * pow(10, BigInt(i))
    }
    
    if isNegative {
        result = -result
    }
    
    words = result.words
}
```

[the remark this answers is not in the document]
rothomp3: Hmm, OK, I see, the `description[...]` will give you the whole string as a `Substring` without making a copy anywhere. OK, I like that, I'll change it.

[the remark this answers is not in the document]
rothomp3: I didn't go with exactly what you suggested, I think mine is actually a little cleaner and also handles the case of `description` being empty *before* the prefix is removed, which I somehow missed completely until now!

rothomp3:
Let me just say I think it's hilarious how much scrutiny the `LosslessStringConvertible` initializer has gotten, all because it happens to be the first method in the file ;)

stephentyrone:
Hi @rothomp3, I created a "biginteger" branch for this work, since there's more development to be done than fits neatly in a single PR. Can you update this PR to point to that branch (I can do this for you if you need help, but it will need to wait a couple days).

rothomp3:
Retargeted PR to `biginteger` branch!

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,752 @@
...
+
+  @inlinable
+  public static func == (lhs: BigInt, rhs: BigInt) -> Bool {
+    lhs.words == rhs.words
+  }
+
+  @inlinable
+  public static func != (lhs: BigInt, rhs: BigInt) -> Bool {
+    !(lhs == rhs)
+  }
```
benrimmington: If you implement `Equatable.==(_:_:)` you should also implement `Hashable.hash(into:)` using the same component(s). Or the compiler can synthesize default implementations for both.

I think the `Equatable.!=(_:_:)` can be removed, because it's the same as the [default implementation](https://github.com/apple/swift/blob/746b58e8e1c1ef1ac09d8031875fd2a08b65597c/stdlib/public/core/Equatable.swift#L179-L197).

benrimmington:
Please could the APIs be grouped and sorted? ([example][])

[example]: <https://gist.github.com/benrimmington/0f2b4356efa38ed239393298260995a7>

Sources/BigInt/BigInt.swift
rothomp3: Technically I believe the auto-generated implementation was already doing exactly this, since `words` is the only stored property on `BigInt`. That said, it rarely hurts to be explicit, so I went ahead and added the `hash(into:)` implementation.

rothomp3:
@benrimmington It's unfortunate that it apparently still requires an approval to be merged into the branch created just for this purpose haha. If this PR were able to be merged there, then you could create a new PR for the reorganization. I'll see about doing it myself, though.

I've been on vacation for the last two weeks or so, but I'm back now!

stephentyrone:
> you could create a new PR for the reorganization

There's no issue with doing this now by creating a PR against your clone.

rothomp3:
That's true, I feel silly for not having thought of that myself.

benrimmington:
Please see rothomp3/swift-numerics#1

benrimmington:
Please see rothomp3/swift-numerics#2

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+      rhsWords.removeLast()
+    }
+
+    if rhsWords.count > lhsWords.count { return (0, lhs) }
+
+    if rhsWords.count == 1 {
+      rhsWords.append(0)
+    }
+
+    func normalize(x: UInt) -> UInt {
```
lorentey: Is there a reason this doesn't just use `leadingZeroBitCount`?

Tests/BigIntTests/BigIntTests.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    let baz = foo + bar
+
+    XCTAssertEqual(baz, BigInt("1234567890123456789112345678901234567890")!)
+
+    let fooz = foo >> BigInt(10)
+    XCTAssertEqual(fooz, foo / 1024)
+
+    let barz = BigInt(1) << 64
+    XCTAssertEqual(barz, BigInt(UInt.max) + 1)
```
lorentey: I think we should have tests covering all of the operations exposed.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public struct BigInt: SignedInteger {
+
+  public typealias Words = [UInt]
+
+  public private(set) var words: Words
```
lorentey: (Implementation tip) For attaswift/BigInt, I found that not allocating a full array for the "small" BigInt cases (up to a couple of digits or so) was a significant overall improvement.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+      }
+    }
+
+    let m = lhsWords.count
+    let n = rhsWords.count
+
+    let bitWidth = UInt(UInt.bitWidth)
+
+    let s = normalize(x: rhsWords[n - 1])
+    let rn = UnsafeMutablePointer<UInt>.allocate(capacity: n)
```
lorentey: Allocating unsafe pointers here seems unnecessary -- wouldn't a temporary array work here?

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+
+    return 1
+  }
+}
+
+// MARK: -
+
+extension BigInt {
+
+  private static func _findQhat(
```
lorentey: It would be a good idea to add an explicit reference to TAOCP so that people can figure out how this works.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+
+extension BigInt: Equatable {
+
+  @inlinable
+  public static func == (lhs: BigInt, rhs: BigInt) -> Bool {
+    lhs.words == rhs.words
+  }
+
+  @inlinable
+  public static func != (lhs: BigInt, rhs: BigInt) -> Bool {
```
lorentey: It isn't necessary to define `!=` here -- it isn't a requirement of `Equatable`.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+
+public struct BigInt: SignedInteger {
+
+  public typealias Words = [UInt]
+
+  public private(set) var words: Words
+
+  @usableFromInline
+  internal init(words: Words) {
+    self.words = words
```
lorentey: I think there is a normalization step missing here -- e.g., `BigInt(words: [0, 0, 0, 0])` should result in an integer with a single digit.

Sources/BigInt/BigInt.swift
rothomp3: Because Knuth didn't have access to that for his book, and I was copying fairly literally! I'll fix that.

Sources/BigInt/BigInt.swift
rothomp3: That is a good idea, yes.

Sources/BigInt/BigInt.swift
rothomp3: The idea with this particular initializer is that it's for internal (really private, but internal is required for `@usableFromInline`) use only, and the words in question should be pre-normalized. So the options here are add a comment to that effect, or risk the slowdown from checking if we need to do normalization here. I'm open to either one, in most cases the check is just going to be false so it doesn't add a huge amount of overhead.

Sources/BigInt/BigInt.swift
rothomp3: In my tests, allocating the memory this way was _significantly_ faster than a temporary array.

Sources/BigInt/BigInt.swift
lorentey: Oh, my apologies, I somehow missed that this is `@usableFromInline` declaration.

The stdlib convention is to put a leading underscore on all internal/private names (whether or not they are also @usableFromInline), to prevent conflicts in case we want to introduce public API with the same name later. (It would be a shame to be forced to choose a suboptimal name because the good name is already taken by an ABI-stable implementation detail.) For initializers, this means putting an underscore on the first argument label, as in `init(_words:)`.

Additionally, we prefer to put `unchecked` (sometimes even `unsafe`) in the name of internal initializers that skip expected normalization/verification steps, to highlight their unusual behavior.

So per the stdlib conventions, the recommended name for this initializer would be something like this:

```swift
 @usableFromInline
  internal init(_uncheckedWords words: Words) { ... }
```

benrimmington: Should there also be a public initializer? For example:

```swift
public init<T>(
  words: T
) where T: RandomAccessCollection, T.Element == UInt, T.Index == Int
```

This would replace the existing `init(bitPattern:)` below (lines 23 to 25), which is unlike the standard library initializers in [IntegerTypes.swift.gyb][].

[IntegerTypes.swift.gyb]: <https://github.com/apple/swift/blob/8510b9234b35938b2869041a9286bd7a0e3bb1c5/stdlib/public/core/IntegerTypes.swift.gyb#L1102-L1115>

Sources/BigInt/BigInt.swift
benrimmington: @lorentey The unit tests will hang (infinite loop?) if:

* the default implementation of `Equatable.!=` is used, or
* the default implementation of `CustomStringConvertible` is used, or
* `String.init(_:radix:uppercase:)` is used with a large `BigInt` value.

The unit tests will crash (stack overflow?) if:

* the default/synthesized implementation of `Equatable.==` is used.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,816 @@
...
+  }
+
+  public init<T>(bitPattern source: T) where T: BinaryInteger {
+    words = Words(source.words)
```
lorentey: There is no requirement that `words` in a BinaryInteger must be minimal -- it may have unnecessary words at the end, especially for (large) FixedWidthIntegers. So the `words` array needs be normalized after initialization here, or things like Equatable/Hashable won't work correctly.

Sources/BigInt/BigInt.swift
lorentey: Definitely! But an initializer like that would also be useful as a `BinaryInteger` requirement, so to prevent source compatibility headaches, we should consider adding it to the stdlib before we add it to SwiftNumerics.

swift-evolution can litigate the exact spelling & constraints, and BigInt can adopt whatever gets accepted.

Sources/BigInt/BigInt.swift
lorentey: That's not ideal -- it sounds like there may be a bad interaction with the generic ==/!= provided by BinaryInteger. We need to understand why it fails & fix it if possible. (It could well be a stdlib issue.)

For `description` and the radix initializer: there may be a bug in the Stdlib's generic radix conversion algorithms for large integers. (Diagnosis and/or patches would be most welcome!)

I don't think BinaryInteger supports `==` synthesis (its generic == will satisfy the Equatable requirement and recursively call itself), so the last item is most likely expected behavior.

rothomp3: So I figured out at least part of this! The default implementation of `!=` apparently calls `<`… which then reveals a rather awful bug in that implementation, to wit… it claims that all integers are less than themselves! Working on a fix now! That said, it _also_ calls `==`, so the explicit implementation of `!=` is better.

Also, why does the default implementation call both `<` and `==`? Shouldn't the latter be sufficient in all cases? I mean, good that it exposed an incredibly bad bug in `BigInt`, but bad if this is happening in the wild and wasting processor cycles haha.

benrimmington: So the [generic `!=`][1] is calling the [generic `==`][2] rather than `BigInt.==` if I understand correctly?

Does the BinaryInteger protocol need to redeclare its inherited Equatable requirements using the `override` keyword?

[1]: <https://github.com/apple/swift/blob/a0b0e362ac3b710af1059ce545aded5ae3b7e5f5/stdlib/public/core/Integers.swift#L1715-L1720>
[2]: <https://github.com/apple/swift/blob/a0b0e362ac3b710af1059ce545aded5ae3b7e5f5/stdlib/public/core/Integers.swift#L1658-L1695>

rothomp3: I'm not sure if it calls the generic `==`, what I confirmed is that it calls _both_ `BigInt.<` _and_ `BigInt.==`. This exposed the nasty bug in `BigInt.<`, which I fixed in the most recent commit. I also left the explicit implementation of `BigInt.!=` in place, since it should be literally twice as fast as calling both `BigInt.<` and `BigInt.==` as the generic `!=` does.

Wait, I figured it out! What it does is call `BigInt.<` every time, then only calls `BigInt.==` if the former returns `false`. Since in the test the number being compared with `0` is always positive, it calls both every time. It is _possible_ that `BigInt.<` will be faster than `BigInt.==`, but the generic `!=` will only be faster in the case that the lhs is in fact less than the rhs. I don't know that it makes sense to optimize for that case, so I favor leaving the explicit `BigInt.!=` in.

benrimmington: You might be able to remove the `BigInt: CustomStringConvertible` extension, now that you've fixed the `<` implementation.

rothomp3: Indeed, done.

lorentey: I wonder if BinaryInteger should provide a non-generic overload of `!==` for the homogeneous case, so that we don't get the call to '<' unless it's actually necessary. (Or we could add a shortcut for the `Other.self == Self.self` case to the generic `BinaryInteger.==`.)

rothomp3: Yeah, I haven't looked at the stdlib code, but if this is happening with say `Int` then it either needs to be fixed or people need to be advised to always phrase inequality comparisons with the expected lower value on the left haha (easy enough when comparing to zero in an unsigned comparison, harder elsewhere)!

benrimmington:
@stephentyrone Will there be an issue with using the same name for a type and its module?

> [Pitch: Fully qualified name syntax](https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482)

Should we add a prefix/suffix to the product and target names? For example:

```swift
import NMXBigInt  // struct BigInt
import NMXComplex // struct Complex
import NMXReal    // protocol Real
```

Sources/BigInt/BigInt.swift
natecook1000: @lorentey It looks like we already provide that as an ambiguity breaker, as `!(lhs == rhs)`. I don't see why that would be calling through to the heterogeneous `==` comparison that uses `<`, though, when `BigInt` defines `==` directly.

rothomp3:
OK, @benrimmington I fixed the failing tests, which had identified a genuine problem! Now we just need to find an implementation of an arbitrarily large `BinaryFloatingPoint` to really test it haha. Theoretically, that would be eligible to be a completely separate addition to this very project, wouldn't it?

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,739 @@
...
+      """)
+
+    let isNegative = source < 0.0
+    var float = isNegative ? -source : source
+
+    if let _ = UInt(exactly: T.greatestFiniteMagnitude) {
+      words = [UInt(float)]
+    } else {
+      var words = Words()
+      let radix = T(sign: .plus, exponent: T.Exponent(UInt.bitWidth), significand: 1)
```
benrimmington: Float16 has a 5-bit exponent (-14 ... 15) so will `T.Exponent(UInt.bitWidth)` be rejected?

Tests/BigIntTests/BigIntTests.swift
```diff
@@ -0,0 +1,479 @@
...
+    000000000000000000000000000000000000000000000000000000000000000000000000000
+    """
+
+  // MARK: - Basic arithmetic
+
+  func testDivision() {
+    let foo = BigInt("12345678901234567890123456789012345678901234567890123456789012345678901234567890")!
+    let bar = BigInt("351235231535161613134135135135")!
+    let baz = foo / bar
+    XCTAssertEqual(baz, BigInt("35149318157164029155780432046477458820396117503007")!)
```
benrimmington: I get a different result with the Python floor-division operator:

```python
foo = 12345678901234567890123456789012345678901234567890123456789012345678901234567890
bar = 351235231535161613134135135135
baz = foo // bar
rem = foo % bar

baz
# 35149318157164029153358504918339691272847595997760

rem
# 15168177833913379478477270290

foo == bar * baz + rem
# True
```

Sources/BigInt/BigInt.swift
rothomp3: It appears it would. There is, AFAIK, currently no Float16 type in Swift; however, there probably will be at some point in the future (it's supported in clang as `_Float16` after all). I've got an experimental branch where I'm implementing one (badly, just for testing) using clang's `_Float16` under the covers, but it's not really working on its own yet haha. Once I get that done, I can at least test this on that branch and figure out a solution (probably to bring back your bit-twiddling solution for `BinaryFloatingPoint` types with bounded `significandBitCount`).

Tests/BigIntTests/BigIntTests.swift
rothomp3: That's worrying. I don't know where I got the expected answer from if it wasn't python! So clearly I'm going to have to write some tests with increasingly large divisions until I find out where it breaks, because the old `description` implementation used the division operator and definitely worked! But it was only ever dividing by 10, so probably there's an issue when the divisor is larger than Int.max? I'll figure it out!

Sources/BigInt/BigInt.swift
rothomp3: Answer is: it works fine! And the main reason is it never hits this code in the Float16 case, because `Float16.greatestFiniteMagnitude` is 65504, which obviously fits exactly in a `UInt`, so the fast path happens. I bet one could prove that any type whose exponent is less than or equal to 6 bits (5 bits in the 32-bit `UInt` case) will always have a `greatestFiniteMagnitude` that fits exactly in `UInt`.

Tests/BigIntTests/BigIntTests.swift
rothomp3: So I'm still working on figuring this out. Division is unquestionably the most complicated part of this whole thing.

rothomp3: Ok, five days later, I have accomplished the fix! It involved re-writing `_findQhat` from scratch. There are now some comments in there pointing to the relevant parts of TAoCP.

rothomp3: oops, there's actually still a bug that current tests don't expose. Fix coming soon, along with hopefully a test to verify it.

rothomp3:
Ok, division now officially works. It probably could be optimized more.

rothomp3:
For example, I could not construct a circumstance where `qhat` was ever larger than `Uint.max + 1`. My guess is this is because of the normalization, and in fact `qhat` will always fit in a single word. If this is true, then `_findQhat` could be simplified quite a bit!

Tests/BigIntTests/BigIntTests.swift
benrimmington: When I add the following to `testDivision()`, there are failures where the `remainder` is negative.

```swift
  func testDivision() {
    for _ in 0 ..< 100 {
      let expected = BigInt(Float64.random(in: 0x1p64 ... 0x1p255))
      let divisor  = BigInt(Float64.random(in: 0x1p64 ... 0x1p128))
      let (quotient, remainder) = expected.quotientAndRemainder(dividingBy: divisor)
      let actual = divisor * quotient + remainder
      XCTAssertEqual(quotient,  expected / divisor)
      XCTAssertEqual(remainder, expected % divisor)
      XCTAssertEqual(
        actual, expected,
        """
        ## FAILURE ##
        ~~~~~~~~~~~~~
              actual: \(actual)
         != expected: \(expected)
        ~~~~~~~~~~~~~
             divisor: \(divisor)
          * quotient: \(quotient)
         + remainder: \(remainder)
        ~~~~~~~~~~~~~
        """)
    }
```

For example:

```text
## FAILURE ##
~~~~~~~~~~~~~
      actual: 34551449761756987820400978628756574120151069926053018399629726511554612953088
 != expected: 34551449761756987820400978628756574120491352292973956863093101118986381164544
~~~~~~~~~~~~~
     divisor: 314741641862945940112684093562865844224
  * quotient: 109777179648832091859813863123394670558
 + remainder: -62781282794539979680630110816414203904
~~~~~~~~~~~~~
```

rothomp3: Ok, I'll take a look. It's almost certainly going to be something really simple like the most significant word of the result being larger than `Int.max`.

rothomp3: Yup, that's what it was.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,799 @@
...
+    self.init(source)
+  }
+
+  public init<T>(truncatingIfNeeded source: T) where T: BinaryInteger {
+    words = Words(source.words)
+  }
+
+  public var bitWidth: Int { words.count * UInt.bitWidth }
+
+  public var trailingZeroBitCount: Int { words.first?.trailingZeroBitCount ?? 0 }
```
nixberg: Shouldn’t that be something like
```Swift
var trailingZeroBitCount: Int {
    var totalZeros = 0
    for word in words {
        let zeros = word.trailingZeroBitCount
        totalZeros += zeros
        if zeros < UInt.bitWidth {
            break
        }
    }
    return totalZeros
}
```

rothomp3: No, imagine that the entire array of `Uint`s is one giant integer. Only the trailing zeros in the least significant word are actually trailing, the rest are in the middle!

natecook1000: If the least significant word is all zeroes (e.g. `UInt.max + 1`), you'll need to keep going up the chain.

rothomp3: Doh! You're right! I feel suddenly stupid, I'll fix that.

rothomp3: Fixed it and added some quick tests for it too.

Sources/BigInt/BigInt.swift
```diff
@@ -395,7 +395,17 @@ extension BigInt: BinaryInteger {
 
   public var bitWidth: Int { words.count * UInt.bitWidth }
 
-  public var trailingZeroBitCount: Int { words.first?.trailingZeroBitCount ?? 0 }
+  public var trailingZeroBitCount: Int {
+    var totalZeros = 0
+    for word in words {
```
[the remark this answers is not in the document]
rothomp3: That's true, I think. I'll take a look.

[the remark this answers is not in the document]
rothomp3: Ok, updated to just explicitly check for `word == 0`, as I agree that's likely to be slightly faster.

Sources/BigInt/BigInt.swift
```diff
@@ -0,0 +1,810 @@
...
+      if !isOverflow && result < Int.max {
+        lhs.words[0] = result
+        return
+      }
+      let knownNegativeResult = lhsWord > Int.max && rhsWord > Int.max
+
+      if lhsWord > Int.max || rhsWord > Int.max, !knownNegativeResult {
+        // positive + negative is always smaller, so overflow is a red herring
+        lhs.words[0] = result
+        return
```
dwaite: Have you evaluated using pattern matches to clarify some of this code and make it a bit easier to spot any edge cases. For instance, here I believe you could do:

```swift

    if lhs.words.count == 1, rhs.words.count == 1 {
      let lhsWord = lhs.words[0]
      let rhsWord = rhs.words[0]
      let lhsNegative = lhsWord > Int.max
      let rhsNegative = rhsWord > Int.max
      let sum = lhsWord &+ rhsWord
      let sumNegative = sum > Int.max
      switch (lhsNegative, rhsNegative, sumNegative) {
        case (true, true, false);
          lhs.words = [sum, UInt.max]
          return
        case (false, false, true);
          lhs.words = [sum, 0]
          return
        default:
          lhs.words[0] = sum
          return
      }
    }
```

rothomp3:
We definitely could do something like that, I’d need to try this code and make sure it actually catches all the cases correctly (looks fine at first glance).

xwu:
Hey @rothomp3, are you still working on this issue? I've been exploring a design where we keep the lowest word inline. It seems promising and I'd like to work on it more, but don't want to step on any ongoing plans you have.

xwu:
Hi all, I've shared an alternative implementation at #120 which starts from a different approach.

rothomp3:
> > Anyway, I actually explored this option myself long before I made this PR, and in my testing it was _dramatically_ worse. It’s possible I’m doing it wrong though!
> 
> Yup. Keeping the lowest word inline sure does make things worse! Check out some of the other ideas I've been exploring over at the other PR though :)

I don't even have a good explanation for _why_, other than perhaps "the people who wrote `Swift.Array` are very smart, and very good at their jobs!"

stephentyrone:
> I don't even have a good explanation for why

The short version is pretty much "special cases make things slower". For some input distributions, it should be a win to have a special small representation (like we do for String), but that's definitely something that we would take about adding down the road--we want to get the general implementation right first.

sadiredd-sv (approved):

ezfe:
What's the status on these changes?

rothomp3:
> What's the status on these changes?

Well, I’m still here, I’m not sure what the next step should be, though!

stephentyrone:
The main thing that's missing at this point is that the tests need to be bulked up pretty enormously. That's something that's OK to do on a branch, though, so I think that it's pretty reasonable to give this a final review and merge it in the near future.

I will probably do a fairly invasive restructuring of the code once it's merged to split it out into somewhat finer-grained files to bring it more in line with the rest of the project, FWIW, which will cause you some hassle if you have changes on other branches that you're waiting to merge after this.

rothomp3:
Nope, go for it!

LiarPrincess:
> The main thing that's missing at this point is that the tests need to be bulked up pretty enormously, and we should be looking at better benchmarks as well. That's something that's OK to do on a branch, though, so I think that it's pretty reasonable to give this a final review and merge it in the near future.

As far as the tests go: there is [Violet - Python VM written in Swift](https://github.com/LiarPrincess/Violet) which has its own implementation of `BigInt`, so this may be useful. Our tests are quite robust, but the problem is that Violet implementation is quite different:
> In Violet we use an union (via [tagged pointer](https://en.wikipedia.org/wiki/Tagged_pointer)) of `Int32` (called `Smi`, after [V8](https://github.com/v8/v8)) and a heap allocation (magnitude + sign representation) with ARC for garbage collection. You can see our [documentation](https://github.com/LiarPrincess/Violet/blob/main/Documentation/BigInt.md) for details.

Which means that technically we have *3 different* implementations of `BigInt`:
- `Bigint` - combines `Smi` and `BigIntHeap` into a single external interface (just [be careful on COW](https://github.com/LiarPrincess/Violet/blob/4fc38c2cdd9fadbc6036407833baa37cfa6654a1/Sources/BigInt/BigInt.swift#L811)) - [BigInt.swift](https://github.com/LiarPrincess/Violet/blob/main/Sources/BigInt/BigInt.swift)
- small `Int32`* stored inline inside the pointer (we call it `Smi`) - [Smi.swift](https://github.com/LiarPrincess/Violet/blob/main/Sources/BigInt/Smi.swift)
- `BigIntHeap` - heap allocated when we do not fit in `Smi` - [BigIntHeap.swift](https://github.com/LiarPrincess/Violet/blob/main/Sources/BigInt/BigIntHeap.swift) for the math part and [BigIntStorage.swift](https://github.com/LiarPrincess/Violet/blob/main/Sources/BigInt/BigIntStorage.swift) for the memory layout/representation

Also, in most of the cases we went with [property based testing](https://en.wikipedia.org/wiki/Property_testing) with means that we test millions of inputs to check if the general rule holds (for example: `a+b=c -> c-a=b` etc.). This takes time, but pays for itself by finding weird overflows in bit operations (we store “sign + magnitude”, so bit operations are a bit difficult to implement).

[*] There is an 64-bit assumption for the whole project. Violet is more-or-less a closed system where we write with the assumption that we are the only consumers. This allows us to concentrate on things that we need, instead of designing for every use case (for `BigInt` we heavily optimize for small integers assuming non-uniform distribution of inputs).

stephentyrone (approved):

LiarPrincess:
What are we doing about tests?

I can create a pull request(s) with tests from [Violet - Python VM written in Swift](https://github.com/LiarPrincess/Violet). In each pull request we would decide if we want them or not. The code has already been written, so this cost was already paid.

If we decide not to merge them then no biggie. I think that it is worth to run them, to check if something fails.

LiarPrincess:
Btw. It may be a good idea to write an 'Int semantics' article for the official Swift blog.

This would include common pitfalls:
- when things overflow - everyone knows that `add`, `sub` and `mul` overflow, but how many programmers know that `div` can also overflow? (signed numbers in 2 complement `Int.min / -1` is unrepresentable)
- rounding modes in shifts - many programmers believe that instead of dividing by a power of 2 you can just shift. Well... this is true in some cases. Tbh. maybe we should expose additional methods with `enum RoundingMode` as an argument?
- etc.

PR #97, opened 2020-01-24
Enable name lookup for modules by suffixing `Module` when needed to break ambiguity.

stephentyrone:
In particular, the following module renames are applied:

- `Real` -> `RealModule`
- `Complex` -> `ComplexModule`

This is a source-breaking change, if you currently directly import these modules. (sorry!)
The fix is to either import `Numerics` instead, or to use the new, suffixed, names. The good news is that fixing this should require only a straightforward search-and-replace for anyone affected.

Some context for *why* I'm making this change can be found [here](https://forums.swift.org/t/real-and-complex-uniform-type-conformance-when-dividing/32536/7). There are a few other use patterns that are enabled by this change, all tied into the same name lookup issue.


rxwei:
Would "Numerics" as a suffix ("ComplexNumerics" and "RealNumerics") read more natural?

stephentyrone:
I think so, but it would be less closely in line with NIO. Worth discussing!

SusanDoggie:
Um... i think putting an underscore is not swifty. I've never seen a project named like this.

stephentyrone:
> Um... i think putting an underscore in the middle is not swifty. I've never seen a project named like this.

https://github.com/apple/swift-standard-library-preview

SusanDoggie:
Is it possible to use "Num" as a prefix?

- NumComplex
- NumReal

xwu:
> (6) is fairly attractive to me, because it's also very clear what's going on, and we could also adopt the rule that you only need to have the suffix when there's a clash to be avoided. So we would have e.g. `RealModule` and `ComplexModule`, but `IntegerUtilities`. Actually, writing this up, I'm talking myself back around to this option.

I like that option; it's like how we have a rule about `Protocol` as a disambiguating suffix, but for modules.

stephentyrone:
> I like that option; it's like how we have a rule about Protocol as a disambiguating suffix, but for modules.

Yeah, this seems like the best approach to me now.

PR #104, opened 2020-03-13
gitignore: ignore vim swap files

compnerd:


stephentyrone:
This seems more like something that should be set in a vim-users local gitignore, but there's no real downside to doing it.

PR #105, opened 2020-03-13
build: add initial CMake based build system

compnerd:
This adds a CMake based build, which enables the build of swift-numerics
on Windows.  This currently does not add the test suite to the build as
there are still some issues to work out in building the full test suite
which require additional math operations to fully build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,29 @@
+#[[
+This source file is part of the Swift Numerics open source project
+
+Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+cmake_minimum_required(VERSION 3.16)
+project(swift-numerics
+  LANGUAGES Swift)
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+
+include(SwiftSupport)
+
+add_subdirectory(Sources)
+
+get_property(SWIFT_NUMERICS_EXPORTS GLOBAL PROPERTY SWIFT_NUMERICS_EXPORTS)
+export(TARGETS ${SWIFT_NUMERICS_EXPORTS}
+  NAMESPACE SwiftNumerics::
+  FILE swift-numerics-config.cmake
+  EXPORT_LINK_INTERFACE_LIBRARIES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,9 +18,13 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
+include(CTest)
 include(SwiftSupport)
 
 add_subdirectory(Sources)
+if(BUILD_TESTING)
+  add_subdirectory(Tests)
+endif()
 
 get_property(SWIFT_NUMERICS_EXPORTS GLOBAL PROPERTY SWIFT_NUMERICS_EXPORTS)
 export(TARGETS ${SWIFT_NUMERICS_EXPORTS}
--- a/Tests/WindowsMain.swift
+++ b/Tests/WindowsMain.swift
@@ -1,0 +1,76 @@
+//===--- WindowsMain.swift ------------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
+// Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if os(Windows)
+import XCTest
+
+@testable
+import RealTests
+
+@testable
+import ComplexTests
+
+extension ElementaryFunctionChecks {
+  static var all = testCase([
+    ("testFloat", ElementaryFunctionChecks.testFloat),
+    ("testDouble", ElementaryFunctionChecks.testDouble),
+  ])
+}
+
+extension IntegerExponentTests {
+  static var all = testCase([
+    ("testFloat", IntegerExponentTests.testFloat),
+    ("testDouble", IntegerExponentTests.testDouble),
+  ])
+}
+
+extension ArithmeticTests {
+  static var all = testCase([
+    ("testPolar", ArithmeticTests.testPolar),
+    ("testBaudinSmith", ArithmeticTests.testBaudinSmith),
+    ("testDivisionByZero", ArithmeticTests.testDivisionByZero),
+  ])
+}
+
+#if canImport(_Differentiation)
+  extension DifferentiableTests {
+    static var all = testCase([
+      ("testComponentGetter", DifferentiableTests.testComponentGetter),
+      ("testConjugate",  DifferentiableTests.testConjugate),
+      ("testArithmetics", DifferentiableTests.testArithmetics),
+    ])
+  }
+#endif
+
+extension PropertyTests {
+  static var all = testCase([
+    ("testProperties", PropertyTests.testProperties),
+    ("testEquatableHashable", PropertyTests.testEquatableHashable),
+    ("testCodable", PropertyTests.testCodable),
+  ])
+}
+
+var testCases = [
+  ElementaryFunctionChecks.all,
+  IntegerExponentTests.all,
+  ArithmeticTests.all,
+  PropertyTests.all,
+]
+
+#if canImport(_Differentiation)
+testCases += [
+  DifferentiableTests.all
+]
+#endif
+
+XCTMain(testCases)
+
+#endif
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -1,0 +1,13 @@
+#[[
+This source file is part of the Swift Numerics open source project
+
+Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+add_subdirectory(_NumericsShims)
+add_subdirectory(ComplexModule)
+add_subdirectory(Numerics)
+add_subdirectory(RealModule)
--- a/Sources/ComplexModule/CMakeLists.txt
+++ b/Sources/ComplexModule/CMakeLists.txt
@@ -1,0 +1,21 @@
+#[[
+This source file is part of the Swift Numerics open source project
+
+Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+add_library(ComplexModule
+  Arithmetic.swift
+  Complex.swift
+  Differentiable.swift)
+set_target_properties(ComplexModule PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
+target_link_libraries(ComplexModule PUBLIC
+  RealModule)
+
+
+_install_target(ComplexModule)
+set_property(GLOBAL APPEND PROPERTY SWIFT_NUMERICS_EXPORTS ComplexModule)
--- a/Sources/Numerics/CMakeLists.txt
+++ b/Sources/Numerics/CMakeLists.txt
@@ -1,0 +1,20 @@
+#[[
+This source file is part of the Swift Numerics open source project
+
+Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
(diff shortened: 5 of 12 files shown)


stephentyrone:
I'm somewhat disinclined to take this, because we don't have any means to test it and it adds an entirely parallel build system that needs to be kept in sync without tests.

compnerd:
Unfortunately, s-p-m doesn't really function on Windows.  The CMake itself does more than s-p-m as it deals with installation as well which is the bulk of the code.  The rest does describe dependencies between targets albeit with explicit source listing.  The glob based builds are actually rather fragile and exhibit odd behaviours as times :(

stephentyrone:
Thanks. What "additional math operations" do we need to build tests?

compnerd:
The additional math operations seem to be related to `__muldc3` and `__mulsc3` IIRC.  I want to try to dig into that and replace them with `_Cmulcc`/`_FCmulcc` if possible.

stephentyrone:
Ah, and probably __div[sd]c3, too. Those are only used as a benchmarking reference, so you can just #if out those test on Windows. I'm planning to move them out of the test target anyway.

stephentyrone:
(That actually sounds like an llvm bug worth fixing; we shouldn't be lowering complex mul/div to a compiler-rt function that doesn't exist on Windows.)

compnerd:
Yeah, I think this is either a LLVM or a Swift issue that needs investigation.  I don't think it makes sense for swift-numerics to special case things - its better to fix it in the compiler.

stephentyrone:
Definitely LLVM. The functions in question are produced by the `*` and `/` operators in the C shims; Swift-Numerics doesn't use them at all, because we have our own multiply and divide implementations.

compnerd:
Oh, thanks for the pointers - should make it easier for me to track down if its just in LLVM

stephentyrone:
Yeah, it should be basically exactly the same type of thing as you fixed for hypotf last year.

compnerd:
@stephentyrone I dont happen to see any shims for that, is this not generated by clang?

stephentyrone:
@compnerd I don't understand the question, but trying to read between the lines: when clang sees a complex multiplication or division (the `*` or `/` operators), it emits a call to `__muldc3` or `__divdc3` to compute the result (after first attempting an inline-computation, in the case of multiply). See: https://godbolt.org/z/zgaBGB

compnerd:
Right, the call generated there is not through the shims but from the compiler lowering the division into a library call.  That’s why I was confused how this could be done through the shims.  I’m just not sure of I’m missing some clever trick.

I recently investigated this a further.  There’s the _Cmulcc and _Cmulcr family of functions on Windows for the complex multiply, however there’s no division routine.  Further more, I’m not sure if the routines are IEEE-754 compliant.  There’s no compiler support for C99 complex in MSVC so I’m a little weary of trying to change clang too aggressively.

What do you think of splitting this work into smaller pieces?  The first change could allow building the library on windows.  I’ll see what we can do on Windows for C99 operations (perhaps we just shim in the clang routines on Windows.  Then I can make subsequent changes to enable building the test suite on windows.

stephentyrone:
> Further more, I’m not sure if the routines are IEEE-754 compliant.

754 says nothing about complex, so no worries there. The applicable standard is C11. We don't really need these for Swift Numerics; they're only used as a baseline for performance measurement. I'll refactor the perf tests into an executable target and the issue will mostly go away. We should still try to fix the bug, but it would no longer be blocking Swift Numerics.

compnerd:
I agree, and I'm willing to go and implement the necessary bits in clang once we have identified what the desired behaviour is.  Currently, our guide (MSVC) does not support this functionality (C11 `_Complex`) so we need to identify the behaviour to implement.

stephentyrone:
If MSVC uses runtime hooks for C++ complex operations, we could investigate using those (after making sure the semantics match with clangs expectations, which I can help with). Or compiler-rt could provide the appropriate symbols on Windows.

compnerd:
I took a look at what happens with C++ complex options.  It seems that it ends up in an open coded division routine:

```cpp
    template <class _Other>
    void _Div(const complex<_Other>& _Right) {
        using _Myctraits = _Ctraits<_Ty>;

        _Ty _Rightreal = static_cast<_Ty>(_Right.real());
        _Ty _Rightimag = static_cast<_Ty>(_Right.imag());

        if (_Myctraits::_Isnan(_Rightreal) || _Myctraits::_Isnan(_Rightimag)) { // set NaN result
            this->_Val[_RE] = _Myctraits::_Nanv();
            this->_Val[_IM] = this->_Val[_RE];
        } else if ((_Rightimag < 0 ? -_Rightimag : +_Rightimag)
                   < (_Rightreal < 0 ? -_Rightreal : +_Rightreal)) { // |_Right.imag()| < |_Right.real()|
            _Ty _Wr = _Rightimag / _Rightreal;
            _Ty _Wd = _Rightreal + _Wr * _Rightimag;

            if (_Myctraits::_Isnan(_Wd) || _Wd == 0) { // set NaN result
                this->_Val[_RE] = _Myctraits::_Nanv();
                this->_Val[_IM] = this->_Val[_RE];
            } else { // compute representable result
                _Ty _Tmp        = (this->_Val[_RE] + this->_Val[_IM] * _Wr) / _Wd;
                this->_Val[_IM] = (this->_Val[_IM] - this->_Val[_RE] * _Wr) / _Wd;
                this->_Val[_RE] = _Tmp;
            }
        } else if (_Rightimag == 0) { // set NaN result
            this->_Val[_RE] = _Myctraits::_Nanv();
            this->_Val[_IM] = this->_Val[_RE];
        } else { // 0 < |_Right.real()| <= |_Right.imag()|
            _Ty _Wr = _Rightreal / _Rightimag;
            _Ty _Wd = _Rightimag + _Wr * _Rightreal;

            if (_Myctraits::_Isnan(_Wd) || _Wd == 0) { // set NaN result
                this->_Val[_RE] = _Myctraits::_Nanv();
                this->_Val[_IM] = this->_Val[_RE];
            } else { // compute representable result
                _Ty _Tmp        = (this->_Val[_RE] * _Wr + this->_Val[_IM]) / _Wd;
                this->_Val[_IM] = (this->_Val[_IM] * _Wr - this->_Val[_RE]) / _Wd;
                this->_Val[_RE] = _Tmp;
            }
        }
    }
```

It would be no worse to just pull in the implementation from compiler-rt I think.

stephentyrone:
This should be just a matter of turning on building those functions for Windows in the compiler-rt build, then, right?

compnerd:
Not really - the thing is that `libclang_rt.builtins-...` is *not* linked against.  That needs to be explicitly found and added to the final link command by the user.  The easiest thing to do really would be to vend that internally in the library sadly.

stephentyrone:
> The easiest thing to do really would be to vend that internally in the library sadly.

In *which* library, though? We don't need these for Swift, so putting them in Swift Numerics or the Swift runtime library makes no sense. We need them for C, and the runtime support library for C (in clang) is compiler-rt.

compnerd:
That is the challenge, `clang` behaves as `cl` and in that case, the runtime support is `vcruntime` which we don't really control.  I agree that disabling `_Complex` in C is definitely drastic.  If we don't want to provide it in the numerics support then I suppose that we can require that the user explicitly adds the library.

stephentyrone:
> If we don't want to provide it in the numerics support

I’m a little confused by this. This is a clang issue, not swift. Compiler-rt *is* the numerics support library for clang.

compnerd:
The problem is that clang is trying to run in a mode where its emulating the behaviour of MSVC intentionally.  That basically means that if you want to link against compiler-rt, you are on your own (as in you need to tell the linker that it needs to be linked against).  Basically, what I am suggesting that we have a few different options:

1. provide the shims in swift-numerics
2. provide the shims in swiftCore
3. create a simple shim that forces auto-linking of `clang_rt.builtins-x86_64.lib`
4. leave it to the user to add the link

Part of the problem is that the builtins are not setup for the MSVC mode, so there will be some more work there as well.  The shims would be easier is all.

stephentyrone:
Right, but options 1 and 2 make no sense, because *Swift doesn't need these operations, C does*. Putting them in swift numerics or swiftCore doesn't do anything to address the problem. Option 3 also makes no sense, because that shim would go ... where? It doesn't matter, because if it existed, it would force auto-linking for C code, which would be the same as just ... linking it for C code.

compnerd:
I think that getting that change into upstream clang would take a while to do since the default behaviour is to be compatible with the de facto platform standard (MSVC).  It may be possible to add a number of flags to control the behaviour in the extension space (and I'm worried that people are going to bring back the long time desire for `-fclang-extensions`).  Im hoping for a more pragmatic solution here.

stephentyrone:
> Im hoping for a more pragmatic solution here.

The pragmatic solution for S-N is "just don't build those benchmarks on Windows," as there's no point in trying to compare ourselves against C functionality that doesn't exist.

compnerd:
@stephentyrone I think that actually makes a lot more sense, thanks for talking through this :)

compnerd:
Okay, I think that we have a good approach for the dependency issue - so the next question is about getting there.  What do you think of splitting up the work, get it building and then do a follow up for the testing?

stephentyrone:
https://github.com/apple/swift-numerics/pull/116

compnerd:
@stephentyrone - okay, tested it locally now.

```
S:\BinaryCache\Release\Windows-x86_64\swift-numerics> S:\BinaryCache\Release\Windows-x86_64\swift-numerics\bin\SwiftNumericsTestRunner.exe
Test Suite 'All tests' started at 2020-06-06 18:16:52.191
Test Suite 'bin.xctest' started at 2020-06-06 18:16:52.207
Test Suite 'ElementaryFunctionChecks' started at 2020-06-06 18:16:52.207
Test Case 'ElementaryFunctionChecks.testFloat' started at 2020-06-06 18:16:52.207
Test Case 'ElementaryFunctionChecks.testFloat' passed (0.0 seconds)
Test Case 'ElementaryFunctionChecks.testDouble' started at 2020-06-06 18:16:52.207
Test Case 'ElementaryFunctionChecks.testDouble' passed (0.0 seconds)
Test Suite 'ElementaryFunctionChecks' passed at 2020-06-06 18:16:52.207
         Executed 2 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
Test Suite 'IntegerExponentTests' started at 2020-06-06 18:16:52.207
Test Case 'IntegerExponentTests.testFloat' started at 2020-06-06 18:16:52.207
Test Case 'IntegerExponentTests.testFloat' passed (0.0 seconds)
Test Case 'IntegerExponentTests.testDouble' started at 2020-06-06 18:16:52.207
Test Case 'IntegerExponentTests.testDouble' passed (0.0 seconds)
Test Suite 'IntegerExponentTests' passed at 2020-06-06 18:16:52.207
         Executed 2 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
Test Suite 'ArithmeticTests' started at 2020-06-06 18:16:52.207
Test Case 'ArithmeticTests.testPolar' started at 2020-06-06 18:16:52.207
Test Case 'ArithmeticTests.testPolar' passed (0.0 seconds)
Test Case 'ArithmeticTests.testBaudinSmith' started at 2020-06-06 18:16:52.207
Test Case 'ArithmeticTests.testBaudinSmith' passed (0.0 seconds)
Test Case 'ArithmeticTests.testDivisionByZero' started at 2020-06-06 18:16:52.207
Test Case 'ArithmeticTests.testDivisionByZero' passed (0.0 seconds)
Test Suite 'ArithmeticTests' passed at 2020-06-06 18:16:52.207
         Executed 3 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
Test Suite 'PropertyTests' started at 2020-06-06 18:16:52.207
Test Case 'PropertyTests.testProperties' started at 2020-06-06 18:16:52.222
Test Case 'PropertyTests.testProperties' passed (0.0 seconds)
Test Case 'PropertyTests.testEquatableHashable' started at 2020-06-06 18:16:52.222
Test Case 'PropertyTests.testEquatableHashable' passed (0.0 seconds)
Test Case 'PropertyTests.testCodable' started at 2020-06-06 18:16:52.222
Test Case 'PropertyTests.testCodable' passed (0.0 seconds)
Test Suite 'PropertyTests' passed at 2020-06-06 18:16:52.222
         Executed 3 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
Test Suite 'bin.xctest' passed at 2020-06-06 18:16:52.222
         Executed 10 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
Test Suite 'All tests' passed at 2020-06-06 18:16:52.222
         Executed 10 tests, with 0 failures (0 unexpected) in 0.0 (0.0) seconds
```

Everything seems good now on Windows!

compnerd:
@stephentyrone - gentle reminder about this :)

Tests/WindowsMain.swift
```diff
@@ -0,0 +1,76 @@
+//===--- LinuxMain.swift --------------------------------------*- swift -*-===//
```
stephentyrone, line 1: File name in comment needs updating.

CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.16)
```
stephentyrone, line 1: The CMake files should have the standard license header as well.

compnerd: Sure, I can put that there, but note that the Swift repository does not follow that convention.

Tests/WindowsMain.swift
compnerd: oops!

compnerd:
@stephentyrone - done, fixed header on the `WindowsMain.swift` and added the copyright headers on the CMakeLists.

CMakeLists.txt
stephentyrone: The Swift repository has a mix of this convention (https://github.com/apple/swift/blob/master/stdlib/public/core/CMakeLists.txt) and omitting them (https://github.com/apple/swift/blob/master/stdlib/public/CMakeLists.txt). I don't think there's a clear precedent.

compnerd: Oh, interesting!  Didn't notice the headers in the stdlib.  Either way, added the headers, it doesn't hurt to have them.

compnerd:
Gentle ping on this.

stephentyrone:
I'm still mildly concerned about the fragility of having a second build system that isn't actually used by most contributors to or users of a project. In particular I would like to have a clear policy about what the responsibilities for contributors testing and updating the CMake build should be. What would you like to see?

stephentyrone:
I think you've resolved any technical concerns that I had, I just want to establish clear guidance for contributors here before landing this.

compnerd:
I completely understand that concern.  I’m planning on adding the ability to build it with build script optionally through CMake.  I think that just updating the list when new files are added would be appreciated, but I do understand that it can break from time to time.  It’s usually pretty easy to repair when noticed (argument parser does the same thing).

It is possible to add a GH actions based build with CMake if there’s interest in that as well.

PR #111, opened 2020-04-19
Make 'Complex' differentiable.

rxwei:
* Conform `Complex` to `Differentiable`.
* Define derivatives for `init(_:_:)`, `real`, `imaginary`, `conjugate`, `+(_:_:)`, `-(_:_:)`, `*(_:_:)`, `/(_:_:)`.

Gated under `#if canImport(_Differentiation)`. Based on some original work done by @bartchr808.

### DIFF
--- a/Sources/ComplexModule/Differentiable.swift
+++ b/Sources/ComplexModule/Differentiable.swift
@@ -1,0 +1,94 @@
+//===--- Differentiable.swift ---------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
+// Copyright (c) 2019 - 2020 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(_Differentiation)
+import _Differentiation
+
+extension Complex: Differentiable
+where RealType: Differentiable, RealType.TangentVector == RealType {
+  public typealias TangentVector = Self
+}
+
+extension Complex
+where RealType: Differentiable, RealType.TangentVector == RealType {
+  @derivative(of: init(_:_:))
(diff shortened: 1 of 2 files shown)


Sources/ComplexModule/Derivatives.swift
```diff
@@ -0,0 +1,94 @@
...
+extension Complex
+where RealType: Differentiable, RealType.TangentVector == RealType {
+  @derivative(of: init(_:_:))
+  @usableFromInline
+  static func _derivativeInit(
+    _ real: RealType,
+    _ imaginary: RealType
+  ) -> (value: Complex, pullback: (Complex) -> (RealType, RealType)) {
+    (value: .init(real, imaginary), pullback: { v in
+      (v.real, v.imaginary)
```
stephentyrone, line 29: Is this definition correct when real or imaginary is `.infinity`?

Sources/ComplexModule/Derivatives.swift
```diff
@@ -0,0 +1,94 @@
...
+  {
+    (lhs * rhs, { v in (rhs * v, lhs * v) })
+  }
+
+  @derivative(of: /)
+  @usableFromInline
+  static func _derivativeDivide(lhs: Complex, rhs: Complex)
+    -> (value: Complex, pullback: (Complex) -> (Complex, Complex))
+  {
+    (lhs / rhs, { v in (v / rhs, -lhs / (rhs * rhs) * v) })
```
stephentyrone, line 81: This expression is prone to spurious overflow and should use the same implementation techniques we use to implement divide itself. In particular, you can't compute `rhs*rhs`.

stephentyrone:
N.B. I'm OK with merging these as-is, so long as we also open bugs to fix the issue and question I flagged.

stephentyrone:
Also merged to swift-5.3 branch.

rahulbhalley:
@rxwei this complex number differentiation is simply wow! Thank you!

PR #118, opened 2020-05-27
[Sketch] Module for Approximate Equality

markuswntr:
# Approximate Equality for Floating Point

This Pull Requests adds `isAlmostEqual(to:tolerance:)` and `isAlmostZero(to:tolerance:)` to `FloatingPoint` as outlined in:

* Issue: #3 
* Proposal: [SE-0259](https://forums.swift.org/t/se-0259-approximate-equality-for-floating-point/23627)

### Account for feedback during the review regarding handling of zero:

Although the proposal has not been accepted, the implementation details from the initial proposal have not been altered in this PR. Based on the feedback received during the review, the desired solution is to unify the handling of `isAlmostEqual` and `isAlmostZero` into a single `isAlmostEqual` function. However, this results in ambiguities that I would like to get feedback on, before changing the implementation in this PR.

Notable examples of possible changes are:

1. Testing for `isAlmostZero` (that is now an **internal** function) in the beginning of `isAlmostEqual` and return early if `self` and `other` are almost zero. `tolerance` is used as absolute tolerance when testing for `isAlmostZero` and is used as relative tolerance for anything else. This has the arguably overall nicest API surface, yet it probably hits on performance for any value that is not (close to) zero – and using `tolerance` as absolute tolerance and relative tolerance is probably undesirable.

```swift
@inlinable
public func isAlmostEqual(
  to other: Self,
  tolerance: Self = Self.ulpOfOne.squareRoot()
) -> Bool {

  // Test if this value is zero, and if the value to compare to is zero
  guard !self.isAlmostZero(absoluteTolerance: tolerance) else {
    return other.isAlmostZero(absoluteTolerance: tolerance)
  }

  assert(tolerance >= .ulpOfOne && tolerance < 1, "tolerance should be in [.ulpOfOne, 1).")

  guard self.isFinite && other.isFinite else {
    return rescaledAlmostEqual(to: other, tolerance: tolerance)
  }

  let scale = max(abs(self), abs(other), .leastNormalMagnitude)
  return abs(self - other) < scale*tolerance
}

@usableFromInline
internal func isAlmostZero(
  absoluteTolerance tolerance: Self // No default
) -> Bool {
  ... // Unchanged
}

@usableFromInline
internal func rescaledAlmostEqual(to other: Self, tolerance: Self) -> Bool {
  ... // Unchanged
}
```

2. Testing for `isAlmostZero` (that is now an **internal** function) in the beginning of `isAlmostEqual` and return early if `self` and `other` are almost zero. The tolerance to test for almost zero is now given by `absoluteTolerance`, while `relativeTolerance` is used for realtive equality tests. This will still probably hit on performance for any value that is not (close to) zero, but it has distinct tolerances for both equality tests. API wise, (IMO) most people will not get tolerances right, as they simply don't know the difference of absolute tolerance and relative tolerance. The default values of both of these parameters may solve the issue though. And arguably, those that need to adjust the tolerance are probably familiar with the differences or can tell them apart from the function header documentation.

```swift
@inlinable
public func isAlmostEqual(
  to other: Self,
  absoluteTolerance: Self = Self.ulpOfOne.squareRoot(), // New
  relativeTolerance: Self = Self.ulpOfOne.squareRoot()  // Renamed
) -> Bool {

  // Test if this value is zero, and if the value to compare to is zero
  guard !self.isAlmostZero(absoluteTolerance: absoluteTolerance) else {
    return other.isAlmostZero(absoluteTolerance: absoluteTolerance)
  }

  assert(relativeTolerance >= .ulpOfOne && relativeTolerance < 1, "relativeTolerance should be in [.ulpOfOne, 1).")

  guard self.isFinite && other.isFinite else {
    return rescaledAlmostEqual(to: other, tolerance: relativeTolerance)
  }

  let scale = max(abs(self), abs(other), .leastNormalMagnitude)
  return abs(self - other) < scale*relativeTolerance
}

@usableFromInline
internal func isAlmostZero(
  absoluteTolerance tolerance: Self // No default
) -> Bool {
  ... // Unchanged
}

@usableFromInline
internal func rescaledAlmostEqual(to other: Self, tolerance: Self) -> Bool {
  ... // Unchanged
}
```

3. Do not change it and test it as is :)


markuswntr:
Hi @stephentyrone!

Thanks for looking into this. This sounds great.
I sure love to look into the proposed changes when the PR is ready.
Let me know if there is anything I can to assist in the meantime.

mseriukov:
@markuswntr Hi! Thank you for this work. 
I'm trying to figure out why the default tolerance is `Self.ulpOfOne.squareRoot()`.
Could you give me some hints, please?

markuswntr:
Hi @mseriukov,
I take no credit. All the work has been done by @stephentyrone.
The discussion on approximate equality of floating points has moved to a new PR #123, where @stephentyrone proposed a unified API design with simpler tolerance parameters and more reasonable default values.
On the subject of `Self.ulpOfOne.squareRoot()`, you can find the details in the initial [Swift Proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0259-approximately-equal.md):

>  /// The default value of `tolerance` is `sqrt(.ulpOfOne)`; this value
>  /// comes from the common numerical analysis wisdom that if you don't
>  /// know anything about a computation, you should assume that roughly
>  /// half the bits may have been lost to rounding. This is generally a
>  /// pretty safe choice of tolerance--if two values that agree to half
>  /// their bits but are not meaningfully almost equal, the computation
>  /// is likely ill-conditioned and should be reformulated.

Issue #35, opened 2019-11-07
Quaternion Math


pashabou:
Since the Numerics modules cover complex numbers and operations, and you mentioned here (#6) that ShapedArrays are within the scope of the project, would quaternion math and possibly other algebras fall under the scope of Swift Numerics as well?

Use cases include everywhere 3D rotation is used, from games and visualization to simulation and control systems. A well-tested and well-documented implementation would be a big boon for a new math library. Some quick googling shows this is not in the C++ STL or a standard part of NumPy, but it is a standard part of MATLAB and part of [Accelerate](https://developer.apple.com/documentation/accelerate/working_with_quaternions) (maybe better suited as part of an update/rewrite of Accelerate?).

stephentyrone:
Hi @pashabou -- quaternions are definitely in-scope, this is a reasonable suggestion.

A couple quick notes:
- a quaternion implementation should wrap the SIMD4 type from the Standard Library
- we'd probably also want to add an explicit 3d vector type, since quaternions are mostly useful as an abstraction for representing rotations and dilations of 3d vector spaces.
- there's an interesting discussion about whether or not it's worth having a unit-quaternion type
- there's an interesting discussion about whether to talk about *quaternions* or *rotations* or both (unit quaternions are a double-cover of rotations, so the two things have different notions of equality).

markuswntr:
Hi @pashabou and @stephentyrone,

I just raised a PR with an implementation sketch for Quaternions.
I hope it is okay for both of you that I linked the PR against this issue (altough I did not open it).

While the PR already lists all the implementation details, I would like to mention, that I am really liking the implementation of the `Complex` type and I tried to align `Quaternion` as much as I could (I even re-used a lot of the comments - sorry).
As both may have similar use cases, I found it reasonable to align them as closly as possible.

I have stripped most of the functionality from my initial implementation sketch down to simply having a PR with only the `Quaternion` type, relevant properties and its most  basic arithmetic. It does not define rotations or vector types. I think it will focus the discussion more on the quaternion definition itself and once we have a consent, to iterate upon it. Though, I intent to provide additional functionality (later).

stephentyrone:
@markuswntr Thanks!

As a heads-up, I'm a little bit swamped for the next few days, but I'll give your PR a quick skim, then provide more detailed feedback sometime next week.

markuswntr:
Hi @stephentyrone!

Thanks for the heads-up. There is no hurry.
And thank you very much for the feedback you already provided.

Issue #12, opened 2019-07-02
Is it desirable to have heterogeneous complex/real operators?


stephentyrone:
I never had heterogeneous real-complex addition or subtraction, and I've [removed](https://github.com/apple/swift-numerics/pull/11) the multiply and division operators because of type-inference issues discussed in the README. This is a minor inconvenience for some use cases, and we should investigate ways in which we could make them available--especially multiplication and division, which
- are used more frequently because of the natural interpretation of C as vector space over R
- would allow us to elide a lot of spurious arithmetic operations an get some perf wins

PR #119, opened 2020-06-04
[Sketch] Quaternion Module

markuswntr:
This PR adds `Quaternion` as part of a new module called `QuaternionModule`.

- Issue: #35

## API Notes

A `Quaternion` contains 4 numeric values conforming to `Real & SIMDScalar`. The numeric type is simply called `RealType` for the remaining notes and considerations. Implementation-wise, `Quaternion` wraps the SIMD4 type from the Standard Library, though it does not expose the SIMD storage publicly – both the initializer and the property (called `components`) are **only** visible to module. It stores the 4 numeric values using a common convention of most graphics libraries in the form of: `xi + yj + zk + r`, i.e. `SIMD4(x, y, z, r)`. 

Publicy, the `Quaternion` is represented by a `real` and an `imaginary` part, where `real` is of `RealType` and `imaginary` is of `SIMD3<RealType>` type. To create a `Quaternion` with matching properties, the following initializers are publicly available:

```swift
init(_ real: RealType, _ imaginary: SIMD3<RealType>)
init(real: RealType, imaginary x: RealType, _ y: RealType, _ z: RealType)
```

Additionally, there are convenience initializer available in the form of:

```swift
init(_ real: RealType) // Equivalent to `Quaternion(real, SIMD3(repeating: 0))`.
init(imaginary: SIMD3<RealType>) // Equivalent to `Quaternion(0, imaginary)`.
init(imaginary x: RealType, _ y: RealType, _ z: RealType) // Equivalent to `Quaternion(real: 0, imaginary: x, y, z)`.
```

Furthermore, `Quaternion` conforms to `ExpressiblyByIntegerLiteral` and `Numeric`,
"therefore" the following initializers are available as well:

```swift
init<Other>(_ real: Other) where Other : BinaryInteger
init?<Other>(exactly real: Other) where Other : BinaryInteger
init(integerLiteral value: Int)
```

And lastly, to allow conversion (and arithmetics) between quaternions of different
floating-point precision, the module exposes the following initializers:

```swift
extension Quaternion where RealType : BinaryFloatingPoint {
  init<Other>(_ other: Quaternion<Other>) where Other : Real & BinaryFloatingPoint & SIMDScalar
  init?<Other>(exactly other: Quaternion<Other>) where Other : Real & BinaryFloatingPoint & SIMDScalar
}
```

The module provides the following static properties on `Quaternion`:

```swift
/// The additive identity, with real and *all* imaginary parts zero.
static var zero: Quaternion { get }
/// The multiplicative identity, with real part one and *all* imaginary parts zero.
static var one: Quaternion { get }
/// The quaternion with the imaginary unit **i** one, i.e. `0 + i + 0j + 0k`.
static var i: Quaternion { get }
/// The quaternion with the imaginary unit **j** one, i.e. `0 + 0i + j + 0k`.
static var j: Quaternion { get }
/// The quaternion with the imaginary unit **k** one, i.e. `0 + 0i + 0j + k`.
static var k: Quaternion { get }
/// The point at infinity.
static var infinity: Quaternion { get }
```

And the following properties:

```swift
/// The conjugate of this quaternion.
var conjugate: Quaternion { get }
/// True if this value is finite.
var isFinite: Bool { get }
/// True if this value is normal.
var isNormal: Bool { get }
/// True if this value is subnormal.
var isSubnormal: Bool { get }
/// True if this value is zero.
var isZero: Bool { get }
/// True if this value is only defined by the imaginary parts.
var isPure: Bool { get }
```

`Quaternion` conforms to `Hashable` (and therefore `Equatable`), `Decodable`,
`Encodable`, `CustomStringConvertible` and `CustomDebugStringConvertible`. The
implementation is rather simple. You can find the details in the `Quaternion.swift` file.

### Arithmetics

A `Quaternion` conforms to `AdditiveArithmetic`. This yields the following arithmetic functions:

```swift
static func + (lhs: Quaternion, rhs: Quaternion) -> Quaternion
static func - (lhs: Quaternion, rhs: Quaternion) -> Quaternion
static func += (lhs: inout Quaternion, rhs: Quaternion)
static func -= (lhs: inout Quaternion, rhs: Quaternion)
```

As well as it conforms to `AlgebraicField` and therefore provides the following arithmetic functions:

```swift
static func * (lhs: Quaternion, rhs: Quaternion) -> Quaternion
static func / (lhs: Quaternion, rhs: Quaternion) -> Quaternion
static func *= (lhs: inout Quaternion, rhs: Quaternion)
static func /= (lhs: inout Quaternion, rhs: Quaternion)
var normalized: Quaternion? { get }
var reciprocal: Quaternion? { get }
```

> Note it does not provide heterogeneous Quaternion/Scalar arithmetic.
> You can find the reason/details in the considerations below.

### Norms

```swift
/// The ∞-norm (`max(abs(a), abs(b), abs(c), abs(d))`).
var magnitude: RealType { get }
/// The Euclidean norm (a.k.a. 2-norm, `sqrt(a*a + b*b + c*c + d*d)`).
var length: RealType { get }
/// The squared length `(a*a + b*b + c*c + d*d)`.
var lengthSquared: RealType { get }
```

While quaternions are often used as an abstraction for representing rotations this PR adds only the most basic type definition and does not define rotation (other than the polar form) or a vector type.
Furthermore, it does not provide a distinct unit quaternion although it provides a normalized variant via the `normalized` property. I think it is reasonable to have a distinct `UnitQuaternion` type that guarantees the underlying quaternion has unity `length`, but I think it should be as-well-as, rather than instead-of.

## Considerations

1. ~~The module provides an initializer to express the imaginary parts of a
quaternion as a tuple of 3 `RealType` values. This somehow "blocks" an initializer
being implemented that simply takes 4 values (e.g. `Quaternion(1,2,3,4)`) but
rather requires the imaginary parts in a additional parentheses (i.e.
`Quaternion(1, (2,3,4))`). The later one was implemented, as the order of the values
within the quaternion (real part first) is immediately visible and it aligns
nicely with the convenience initializer that takes imaginary values only as a tuple
(i.e. `Quaternion(imaginary: (2,3,4))`) but I have no strong opinion on it and
would love to get your feedback.~~ This has since been changed to: `init(real: RealType, imaginary x: RealType, _ y: RealType, _z: RealType)`
2. The `Quaternion` wraps the `SIMD4` type, however its storage is not exposed
publicly (nor is the initializer) as all of the necessary operations around
quaternions should (later) be available on the `Quaternion`, or via its
`real` and `imaginary` properties. This may yield performance issues for applications
where the entire storage is required. However it guarantees that all values have
fixup the semantics for non-finite values (much like `x/y` and `real/imaginary`
for `Complex` types).
3. The expressions `q * 1`, where `q` is a quaternion, will be interpreted by the
compiler as `q * Quaternion(1, (0,0,0))` due to the `ExpressibleByIntegerLiteral`
conformance and the implementation details of it. This aligns nicely with `q * .one`,
as the multiplicative identity of a quaternion (or `one`) is defined as
`Quaternion(1, (0,0,0))`. However, this adds ambiguity when viewed as vector
space structure, where `q * 1` should probably yield `q * Quaternion(1, (1,1,1))`.
This is the main reason why there is no heterogenous arithmetic available, much
like it is not available on the `Complex` type (Issue #12). I think, should it
ever "be fixed", it should be implemented on both types at the same time and equally.
4. ~~`Quaternion` defines `static var i: Quaternion` to return a pure quaternion with
all imaginary components at 1, i.e. `Quaternion(0, (1,1,1))`. This may be misleading
tough, as a quaternion is commonly written in the form of `a + bi + cj + dk` and could imply
that only the second (b**i**) part is 1. However, naming the static property
`imaginary` (to make it clearer that **all** imaginary components are of interest)
makes it so that there is a static property `static var imaginary: Quaternion` and a
non-static property `var imaginary: SIMD3<RealType>` available on quaternion, which
is also misleading. One solution to this could be to add `static var j: Quaternion`
and `static var k: Quaternion`. So to get a new quaternion with only the imaginary
components of another quaternion one could simply write `q * (.i + .j + .k)`.~~ There are now three distinct static properties available, named: `i`, `j` and `k`
5. Right now, the `Quaternion` does not provide a polar representation,
tough it should become part of this PR. Main reason: I would like to get your
opinion on it. Right now, I think the best way to define a polar form is via
an `angle` of `RealType` around an `axis` of `SIMD3<RealType` – much like it is defined
for [`simd_quatf`][simd_quatf] and [`simd_quatd`][simd_quatd].
However, to match the `Complex` type, it should probably be `axis` first and `angle` last.
This would add the following API to `Quaternion`:

```swift
init(axis: SIMD3<RealType>, phase: RealType)
var axis: SIMD3<RealType> { get }
var angle: RealType { get }
var polar: (axis: SIMD3<RealType>, angle: RealType) {
  (axis, angle)
}
```

## Missing

- [ ] Polar Form
- [ ] Additional Tests around arithmetics
- [ ] Additional Documentation

[simd_quatf]: https://developer.apple.com/documentation/simd/simd_quatf/2914611-init
[simd_quatd]:https://developer.apple.com/documentation/simd/simd_quatd/2914986-init

### DIFF
--- a/Sources/QuaternionModule/Quaternion.swift
+++ b/Sources/QuaternionModule/Quaternion.swift
@@ -1,0 +1,499 @@
+//===--- Quaternion.swift -------------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
+// Copyright (c) 2019 - 2020 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import RealModule
+
+/// A quaternion represented by a real and three imaginary parts.
+///
+/// TODO: More informations on type
+///
+/// Implementation notes:
+/// -
+///
+/// `.magnitude` does not return the Euclidean norm; it uses the "infinity
+/// norm" (`max(|a|,|b|,|c|,|d|)`) instead. There are two reasons for this
+/// choice: first, it's simply faster to compute on most hardware. Second,
+/// there exist values for which the Euclidean norm cannot be represented.
+/// Using the infinity norm avoids this problem entirely without significant
+/// downsides. You can access the Euclidean norm using the `length` property.
+/// See `Complex` type of the swift-numerics package for additional details.
+public struct Quaternion<RealType> where RealType: Real & SIMDScalar {
+
+  /// The components of the 4-dimensional vector space of the quaternion.
+  ///
+  /// Components are stored within a 4-dimensional SIMD vector with the scalar component
+  /// first, i.e. representing the most common mathmatical representation that is:
+  ///
+  ///     a + bi + cj + dk
+  @usableFromInline @inline(__always)
+  internal var components: SIMD4<RealType>
+
+  /// Creates a new quaternion from given 4-dimensional vector.
+  ///
+  /// This initializer creates a new quaternion by reading the values of the vector as components
+  /// of the quaternion with the scalar component ordered first, i.e in the form of:
+  ///
+  ///     a + bi + cj + dk
+  ///
+  /// - Parameter components: The components of the 4-dimensionsal vector space of the quaternion,
+  /// scalar part first.
+  @_transparent
+  public init(from components: SIMD4<RealType>) {
+    self.components = components
+  }
+}
+
+// MARK: - Basic Property
+extension Quaternion {
+  /// The real part of this quaternion value.
+  public var real: RealType {
+    @_transparent
+    _read { yield components[0] }
+
+    @_transparent
+    _modify { yield &components[0] }
+  }
+
+  /// The imaginary part of this quaternion value.
+  public var imaginary: SIMD3<RealType> {
+    @_transparent
+    get { components[SIMD3(1,2,3)] }
+
+    @_transparent
+    set {
+        components[1] = newValue[0]
+        components[2] = newValue[1]
+        components[3] = newValue[2]
+    }
+  }
+
+  /// The additive identity, with real and imaginary parts all zero.
+  ///
+  /// See also:
+  /// -
+  /// - .one
+  /// - .i
+  /// - .infinity
+  @_transparent
+  public static var zero: Quaternion {
+    .init(0)
+  }
+
+  /// The multiplicative identity, with real part one and imaginary parts all zero.
+  ///
+  /// See also:
+  /// -
+  /// - .zero
+  /// - .i
+  /// - .infinity
+  @_transparent
+  public static var one: Quaternion {
+    .init(1)
+  }
+
+  /// The imaginary unit.
+  ///
+  /// See also:
+  /// -
+  /// - .zero
+  /// - .one
+  /// - .infinity
+  @_transparent
+  public static var i: Quaternion {
+    .init(imaginary: SIMD3(repeating: 1))
+  }
+
+  /// The point at infinity.
+  ///
+  /// See also:
+  /// -
+  /// - .zero
+  /// - .one
+  /// - .i
+  @_transparent
+  public static var infinity: Quaternion {
+    .init(.infinity)
+  }
+
+  /// The conjugate of this quaternion value.
+  @_transparent
+  public var conjugate: Quaternion {
+    .init(from: components.replacing(with: -components, where: [false, true, true, true]))
+  }
+
+  /// True if this value is finite.
+  ///
+  /// A quaternion value is finite if neither component is infinity or nan.
+  ///
+  /// See also:
+  /// -
+  /// - `.isNormal`
+  /// - `.isSubnormal`
+  /// - `.isZero`
+  /// - `.isPure`
+  @_transparent
+  public var isFinite: Bool {
+    components.x.isFinite
+        && components.y.isFinite
+        && components.z.isFinite
+        && components.w.isFinite
+  }
+
+  /// True if this value is normal.
+  ///
+  /// A quaternion is normal if it is finite and *either* the real or imaginary component is normal.
+  /// A floating-point number representing one of the components is normal if its exponent allows a
+  /// full-precision representation.
+  ///
+  /// See also:
+  /// -
+  /// - `.isFinite`
+  /// - `.isSubnormal`
+  /// - `.isZero`
+  /// - `.isPure`
+  @_transparent
+  public var isNormal: Bool {
+    isFinite && (
+        real.isNormal || (imaginary.x.isNormal && imaginary.y.isNormal && imaginary.z.isNormal)
+    )
+  }
+
+  /// True if this value is subnormal.
+  ///
+  /// A quaternion is subnormal if it is finite, not normal, and not zero. When the result of a
+  /// computation is subnormal, underflow has occurred and the result generally does not have full
+  /// precision.
+  ///
+  /// See also:
+  /// -
+  /// - `.isFinite`
+  /// - `.isNormal`
+  /// - `.isZero`
+  /// - `.isPure`
+  @_transparent
+  public var isSubnormal: Bool {
+    isFinite && !isNormal && !isZero
+  }
+
+  /// True if this value is zero.
+  ///
+  /// A quaternion is zero if *both* the real and imaginary components are zero.
+  ///
+  /// See also:
+  /// -
+  /// - `.isFinite`
+  /// - `.isNormal`
+  /// - `.isSubnormal`
+  /// - `.isPure`
+  @_transparent
+  public var isZero: Bool {
+    return components.x.isZero
+        && components.y.isZero
+        && components.z.isZero
+        && components.w.isZero
+  }
+
+  /// True if this value is only defined by the imaginary part (`real == .zero`)
+  @_transparent
+  public var isPure: Bool {
+    real.isZero
+  }
+
+  /// The ∞-norm of the value (`max(abs(real), abs(imaginary))`).
+  ///
+  /// If you need the Euclidean norm (a.k.a. 2-norm) use the `length` or `lengthSquared`
+  /// properties instead.
+  ///
+  /// Edge cases:
+  /// -
+  /// - If `z` is not finite, `z.magnitude` is `.infinity`.
+  /// - If `z` is zero, `z.magnitude` is `0`.
+  /// - Otherwise, `z.magnitude` is finite and non-zero.
+  ///
+  /// See also:
+  /// -
+  /// - `.length`
+  /// - `.lengthSquared`
+  @_transparent
+  public var magnitude: RealType {
+    guard isFinite else { return .infinity }
+    return max(abs(components.max()), abs(components.min()))
+  }
+}
+
+// MARK: - Additional Initializers
+extension Quaternion {
+  /// The quaternion with specified real part and zero imaginary part.
+  ///
+  /// Equivalent to `Quaternion(real, SIMD3(repeating: 0))`.
+  @inlinable
+  public init(_ real: RealType) {
+    self.init(real, SIMD3(repeating: 0))
+  }
+
+  /// The quaternion with specified imaginary part and zero real part.
+  ///
+  /// Equivalent to `Quaternion(0, imaginary)`.
+  @inlinable
+  public init(imaginary: SIMD3<RealType>) {
+    self.init(0, imaginary)
+  }
+
+  /// The quaternion with specified imaginary part and zero real part.
+  ///
+  /// Equivalent to `Quaternion(0, imaginary)`.
+  @inlinable
+  public init(imaginary: (b: RealType, c: RealType, d: RealType)) {
+    self.init(imaginary: SIMD3(imaginary.b, imaginary.c, imaginary.d))
+  }
+
+  /// The quaternion with specified real part and imaginary parts.
+  @inlinable
+  public init(_ real: RealType, _ imaginary: SIMD3<RealType>) {
+    self.init(from: SIMD4(real, imaginary.x, imaginary.y, imaginary.z))
+  }
+
+  /// The quaternion with specified real part and imaginary parts.
+  @inlinable
+  public init(_ real: RealType, _ imaginary: (b: RealType, c: RealType, d: RealType)) {
+    self.init(real, SIMD3(imaginary.b, imaginary.c, imaginary.d))
+  }
+
+  /// The quaternion with specified real part and zero imaginary part.
+  ///
+  /// Equivalent to `Quaternion(RealType(real))`.
+  @inlinable
+  public init<Other: BinaryInteger>(_ real: Other) {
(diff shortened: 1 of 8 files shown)


Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+public struct Quaternion<RealType> where RealType: Real & SIMDScalar {
+
+  /// The components of the 4-dimensional vector space of the quaternion.
+  ///
+  /// Components are stored within a 4-dimensional SIMD vector with the scalar part
+  /// first, i.e. representing the most common mathmatical representation that is:
+  ///
+  ///     a + bi + cj + dk
+  @usableFromInline @inline(__always)
+  internal var components: SIMD4<RealType>
```
stephentyrone, line 44: I *think* I might prefer to simply have `(x, y, z, w)` instead of `SIMD4<RealType>` for storage, so that we can drop the `SIMDScalar` requirement. It definitely makes some of the implementations a little nicer to write, but doesn't benefit the surface API very much, so I'm not sure it's worth the extra requirement. I am not 100% certain about this, however.

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+  /// - .i
+  @_transparent
+  public static var infinity: Quaternion {
+    Quaternion(.infinity)
+  }
+
+  /// The conjugate of this quaternion.
+  @_transparent
+  public var conjugate: Quaternion {
+    Quaternion(from: components.replacing(with: -components, where: [false, true, true, true]))
```
stephentyrone, line 135: I would probably write this as Quaternion(from: components * [1, -1, -1, -1]) (or [-1, -1, -1, 1] if we flip the storage order as I mentioned before).

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+  /// See also:
+  /// -
+  /// - `.isFinite`
+  /// - `.isSubnormal`
+  /// - `.isZero`
+  /// - `.isPure`
+  @_transparent
+  public var isNormal: Bool {
+    let realIsNormal = components.x.isNormal
+    let imaginaryIsNormal = components.y.isNormal && components.z.isNormal && components.w.isNormal
```
stephentyrone, line 171: This should be normal if _any_of the components is normal, not all of them.

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+  ///
+  /// See also:
+  /// -
+  /// - `.isFinite`
+  /// - `.isNormal`
+  /// - `.isSubnormal`
+  /// - `.isPure`
+  @_transparent
+  public var isZero: Bool {
+    return components.x.isZero
```
stephentyrone, line 204: You can write this components == .zero.

stephentyrone:
> The module provides an initializer to express the imaginary parts of a
quaternion as a tuple of 3 RealType values. This somehow "blocks" an initializer
being implemented that simply takes 4 values (e.g. Quaternion(1,2,3,4)) but
rather requires the imaginary parts in a additional parentheses (i.e.
Quaternion(1, (2,3,4))). The later one was implemented, as the order of the values
within the quaternion (real part first) is immediately visible and it aligns
nicely with the convenience initializer that takes imaginary values only as a tuple
(i.e. Quaternion(imaginary: (2,3,4))) but I have no strong opinion on it and
would love to get your feedback.

From an ergonomics and clarity standpoint, it might be nice to do:
```swift
public init(real: RealType, imaginary x: RealType, _ y: RealType, _ z: RealType)
```
This is pretty weird at the declaration site, but in use it gives us: `Quaternion(real: 1, imaginary: 0, 1, 0)`, which reads quite well.

stephentyrone:
> Quaternion defines static var i: Quaternion to return a pure quaternion with
all imaginary components at 1, i.e. Quaternion(0, (1,1,1)). This may be misleading
tough, as a quaternion is commonly written in the form of a + bi + cj + dk and could imply
that only the second (bi) part is 1. However, naming the static property
imaginary (to make it clearer that all imaginary components are of interest)
makes it so that there is a static property static var imaginary: Quaternion and a
non-static property var imaginary: SIMD3<RealType> available on quaternion, which
is also misleading. One solution to this could be to add static var j: Quaternion
and static var k: Quaternion. So to get a new quaternion with only the imaginary
components of another quaternion one could simply write q * (.i + .j + .k).

Yeah, I would prefer to have `.i`, `.j`, and `.k`. We can make up some other API for this if we need it, but let's just use them for now.

stephentyrone:
> Right now, the Quaternion does not provide a polar representation,
tough it should become part of this PR. Main reason: I would like to get your
opinion on it. Right now, I think the best way to define a polar form is via
an angle of RealType around an axis of SIMD3<RealType – much like it is defined
for simd_quatf and simd_quatd.

Strictly speaking, this might construct a hypothetical UnitQuaternion type; one option would be to have a `length` parameter as well to construct a (non-unit) Quaternion?

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+  ///     a + bi + cj + dk
+  @usableFromInline @inline(__always)
+  internal var components: SIMD4<RealType>
+
+  /// A quaternion constructed from given 4-dimensional SIMD vector.
+  ///
+  /// Creates a new quaternion by reading the values of the SIMD vector
+  /// as components of a quaternion with the scalar part first, i.e. in the form of:
+  ///
+  ///     a + bi + cj + dk
```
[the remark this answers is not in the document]
markuswntr: Thanks! The order is now flipped to (x,y,z,r) – but for the internal storage only. It remains (r,x,y,z) for the public API, i.e. all the initializers.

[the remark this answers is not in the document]
markuswntr: I also renamed all the commentary using the notation `a + bi + cj + dk` to `r + xi + zj + zk` in the process. It aligns better with the storage and the initializer.

Sources/QuaternionModule/Quaternion.swift
markuswntr: This is definitely easier to read/understand, thanks!

Sources/QuaternionModule/Quaternion.swift
markuswntr: My bad, thank you!

Sources/QuaternionModule/Quaternion.swift
markuswntr: I changed it, thanks!

markuswntr:
> From an ergonomics and clarity standpoint, it might be nice to do:
> 
> ```swift
> public init(real: RealType, imaginary x: RealType, _ y: RealType, _ z: RealType)
> ```
> 
> This is pretty weird at the declaration site, but in use it gives us: `Quaternion(real: 1, imaginary: 0, 1, 0)`, which reads quite well.

I definitely like this initializer having named properties. I ditched the idea while writing the module, as it ends up being quit long and I was afraid others might prefer a shorter initializer – and just put 4 values in order into it (hence the imaginary tuple, to separate real from imaginary somehow). But, as you mentioned, from an ergonomics and clarity standpoint, it is just nicer, so I'll change it to your suggestion – if necessary, I guess we could later iterate on it.

stephentyrone:
Markus, I think that this is a reasonable enough start that we should land it on a branch and iterate there. Can you retarget your PR against https://github.com/apple/swift-numerics/tree/Quaternions?

markuswntr:
I have retargeted the branch, thank you.
For completeness I have updated the notes on the PR to reflect the latest changes.

stephentyrone:
Excellent, thanks Markus. I'm going to take this as is, since I think the basic design is sound, and we can iterate on it with smaller PRs for a while before merging to master.

Sources/QuaternionModule/Quaternion.swift
markuswntr: I do agree on the `SIMDScalar` requirement. I definitely think this is an important issue to talk about and to maybe iterate a little longer – and I don't mind rewriting it. I'll raise a PR solely for this topic, so we can have a discussion of whether or not it is worth having an SIMD storage and the inevitable `SIMDScalar` requirement – while also seeing the benefits and disadvantages it implies on the API and implementations.

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+        let d = RealType(exactly: other.components.w)
+    else { return nil }
+    self.init(from: SIMD4(a, b, c, d))
+  }
+}
+
+// MARK: - Conformance to Hashable and Equatable
+extension Quaternion: Hashable {
+  @_transparent
+  public static func == (lhs: Quaternion, rhs: Quaternion) -> Bool {
```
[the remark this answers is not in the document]
markuswntr: Yes, you are absolute right – thank you!
I'll try to document that clearly in one of the next PRs.

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -0,0 +1,365 @@
...
+    guard lhs.isFinite || rhs.isFinite else { return true }
+    // For finite numbers, equality is defined componentwise. Cases where
+    // only one of lhs or rhs is infinite fall through to here as well, but this
+    // expression correctly returns false for them so we don't need to handle
+    // them explicitly.
+    return lhs.components == rhs.components
+  }
+
+  @_transparent
+  public func hash(into hasher: inout Hasher) {
```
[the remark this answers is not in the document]
markuswntr: I did some tinkering and, while I am still very uncertain, I was thinking if this is maybe something that we could move to a specific `UnitQuaternion` (or `Rotation3`, or ...) – so that we'll avoid collisions on this quaternion? This would allow us to also alter the `Equatable` implementation for `UnitQuaternion` so that it will return `true` for `x == -x`.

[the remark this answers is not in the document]
markuswntr: That said, I'll raise a PR incorporating your feedback to make x and -x hash to the same value as I am very uncertain about my suggestion.

markuswntr:
Hi Stephen,

while gathering the TODOs from the PR I realized, I have not mentioned
the issue with the algebraic field conformance on Quaternion.

Quaternions do not form a commutative group under its multiplication – their multiplication is non-commutative.
Dropping the condition that multiplication is commutative leads to a division ring rather than a field.

So strictly speaking, conforming to `AlgebraicField` is incorrect and I wanted to raise a question if
this is maybe something to change in the overall protocol structure of the `RealModule`.
Is this something that should be addressed within a new GitHub issue for now?
Or should the `AlgebraicField` conformance be dropped for now?

I am very sorry. This should have been part in the considerations of the PR.

alexanderuv:
@stephentyrone ☝️ maybe you didn't see that
Thanks @markuswntr for the implementation. This will be very useful as mentioned before for 3D related programming!

PR #120, opened 2020-06-08
[Sketch] BigInt

xwu:
Here's another stab at implementing a `BigInt` type.

## Implementation

I started with a sign-magnitude representation, then made a few gradual refinements (sadly, the Git history got nuked somewhere in there):

- Taking inspiration from the Java implementation, I switched to storing the _signum_ instead of the _sign_; this allows us to eliminate the two possible representations of zero.

- I decided to include the count of trailing zero words (i.e., an exponent) in a "combination" field, where `_combination = _signum * (_exponent + 1)`. This not only permits constant-time random access of the two's complement representation of negative values, but allows us not to store those trailing zeros. To determine whether a value is a power of two, for instance, it suffices to check whether there is only one nonzero word that has only one nonzero bit.

The resultant representation is somewhat reminiscent of IEEE floating point, with each value decomposed into a signum, exponent, and significand such that notionally `value = _signum * _significand << (UInt.bitWidth * _exponent)`. Overall, it seems to meld the best of both worlds (sign-magnitude and two's complement) for the reasons given above. I don't know of another implementation of `BigInt` that takes this approach, so either it's something new and worthwhile or a very silly diversion.

Taking @lorentey's point that it is best to have some accommodation for storing smaller numbers without allocating an array, I take a slightly different approach here: instead of storing two words in a tuple and any longer values in an array, this implementation unconditionally stores the low word of the significand inline. In this way, serially incrementing or decrementing even most large values should not trigger a copy-on-write operation.

Both schoolbook and Karatsuba multiplication are implemented; currently, the cutoff is at hardcoded at 16 words. Bitwise operators perform as they should (working with the logical two's complement representation—in fact, for simplicity, their implementations make use of efficient constant-time random access to the actual two's complement representation that is provided by `Words`).

Currently, no new APIs that aren't already present on standard library integer types are introduced, other than the initializer `init(words:)` to create a value from a list of `UInt`-typed words.

### To-dos:
[ ] Implement division
[ ] Implement the most salient `BigInt`-specific operations (can be done in a follow-up PR).

## Testing

I have verified that operations currently implemented (all but division, that is) produce the expected results. Not all edge cases are covered yet, however.

So that I can test calculations using randomly generated operands, I use `attaswift/BigInt` as a reference implementation and check that the current implementation produces the same result as does `attaswift/BigInt` (with one exception, where on manual checking it appears that the latter implementation has a bug).

### To-dos:
[ ] Test random number generation APIs
[ ] Compare and optimize performance (can be done in a follow-up PR).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,18 +15,22 @@ import PackageDescription
 let package = Package(
   name: "swift-numerics",
   products: [
+    .library(name: "BigIntModule", targets: ["BigIntModule"]),
     .library(name: "ComplexModule", targets: ["ComplexModule"]),
     .library(name: "Numerics", targets: ["Numerics"]),
     .library(name: "RealModule", targets: ["RealModule"]),
   ],
   dependencies: [
+    .package(url: "https://github.com/attaswift/BigInt.git", from: "3.0.0")
   ],
   targets: [
+    .target(name: "BigIntModule", dependencies: []),
     .target(name: "ComplexModule", dependencies: ["RealModule"]),
     .target(name: "Numerics", dependencies: ["ComplexModule", "RealModule"]),
     .target(name: "_NumericsShims", dependencies: []),
     .target(name: "RealModule", dependencies: ["_NumericsShims"]),
     
+    .testTarget(name: "BigIntTests", dependencies: ["BigIntModule", "BigInt"]),
     .testTarget(name: "ComplexTests", dependencies: ["ComplexModule", "_NumericsShims"]),
     .testTarget(name: "RealTests", dependencies: ["RealModule"]),
   ]
--- a/Sources/BigIntModule/BigInt.Words.swift
+++ b/Sources/BigIntModule/BigInt.Words.swift
@@ -1,0 +1,76 @@
+extension BigInt {
+  /// The words of an arbitrarily large signed integer.
+  ///
+  /// For a negative value, words are of the two’s complement representation.
+  @frozen
+  public struct Words {
+    /// The arbitrarily large signed integer.
+    @usableFromInline
+    internal var _value: BigInt
+
+    /// Creates a `BigInt.Words` from the given arbitrarily large signed
+    /// integer.
+    ///
+    /// - Parameter value: The arbitrarily large signed integer.
+    @inlinable
+    public init(_ value: BigInt) {
+      _value = value
+    }
+  }
+}
+
+extension BigInt.Words: RandomAccessCollection {
+  @inlinable
+  public var count: Int {
+    if _value._combination == 0 { return 1 }
+    let temporary = _value._exponent + _value._significand.count
+    let lastIndex = _value._significand.count &- 1
+    let highWord = _value._significand[lastIndex]
+    guard Int(bitPattern: highWord) < 0 else {
+      return temporary
+    }
+    // If the leading bit is set, then--
+    //
+    // For a positive value:
+    // We need to add at least one leading zero bit (and therefore one
+    // additional word) for a signed representation.
+    if _value._combination > 0 { return temporary + 1 }
+    // For a negative value:
+    // The two's complement of a magnitude has the same bit width as that of
+    // the magnitude itself if and only if the magnitude is a power of two.
+    // Otherwise, we need one additional bit (and therefore one additional
+    // word) to fit the two's complement.
+    //
+    // (Note that `(x & (x &- 1)) == 0` is a method of determining if `x` is
+    // a power of two.)
+    return lastIndex == 0 && (highWord & (highWord &- 1)) == 0
+      ? temporary
+      : temporary + 1
+  }
+  
+  @inlinable
+  public var startIndex: Int { 0 }
+  
+  @inlinable
+  public var endIndex: Int { count }
+  
+  @inlinable
+  public func index(before i: Int) -> Int { i - 1 }
+  
+  @inlinable
+  public func index(after i: Int) -> Int { i + 1 }
+  
+  @inlinable
+  public subscript(position: Int) -> UInt {
+    precondition(position >= 0, "Index out of bounds")
+    guard position >= _value._exponent else { return 0 }
+    let idx = position &- _value._exponent
+    guard idx < _value._significand.count else {
+      return _value._combination < 0 ? UInt.max : 0
+    }
+    let word = _value._significand[idx]
+    return _value._combination < 0
+      ? idx == 0 ? ~word &+ 1 : ~word
+      : word
+  }
+}
--- a/Sources/BigIntModule/BigInt._Significand.swift
+++ b/Sources/BigIntModule/BigInt._Significand.swift
@@ -1,0 +1,663 @@
+extension BigInt {
+  /// The significand of a `BigInt` value, a nonempty collection of the
+  /// significant digits of that value's magnitude stored in words of type
+  /// `UInt`.
+  ///
+  /// The first element of the collection is the lowest word.
+  @frozen
+  @usableFromInline
+  internal struct _Significand {
+    /// The low word of this significand.
+    @usableFromInline
+    internal var _low: UInt
+    
+    /// The rest of the words of this significand, from second lowest to
+    /// highest.
+    @usableFromInline
+    internal var _rest: [UInt]
+    
+    /// Creates a new significand with a single low word equal to zero.
+    @inlinable
+    internal init() {
+      _low = 0
+      _rest = []
+    }
+
+    /// Creates a new significand with the given words.
+    @inlinable
+    internal init(_ low: UInt, _ rest: [UInt] = []) {
+      _low = low
+      _rest = rest
+    }
+  }
+}
+
+extension BigInt._Significand: RandomAccessCollection, MutableCollection {
+  @inlinable
+  internal var count: Int { _rest.count + 1 }
+  
+  @inlinable
+  internal var startIndex: Int { 0 }
+  
+  @inlinable
+  internal var endIndex: Int { count }
+  
+  @inlinable
+  internal func index(before i: Int) -> Int { i - 1 }
+  
+  @inlinable
+  internal func index(after i: Int) -> Int { i + 1 }
+  
+  @inlinable
+  internal subscript(position: Int) -> UInt {
+    _read {
+      if position == 0 {
+        yield _low
+      } else {
+        yield _rest[position - 1]
+      }
+    }
+    _modify {
+      if position == 0 {
+        yield &_low
+      } else {
+        yield &_rest[position - 1]
+      }
+    }
+  }
+}
+
+/// `BigInt._Coefficient` can never be empty, so it cannot conform to
+/// `RangeReplaceableCollection`; however, we will implement some of its
+/// requirements here.
+///
+/// Documentation is transposed from the corresponding methods on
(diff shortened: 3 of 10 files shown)


Package.swift
```diff
@@ -15,18 +15,22 @@ import PackageDescription
...
     .library(name: "Numerics", targets: ["Numerics"]),
     .library(name: "RealModule", targets: ["RealModule"]),
   ],
   dependencies: [
+    .package(url: "https://github.com/attaswift/BigInt.git", from: "3.0.0")
   ],
   targets: [
+    .target(name: "BigIntModule", dependencies: []),
     .target(name: "ComplexModule", dependencies: ["RealModule"]),
     .target(name: "Numerics", dependencies: ["ComplexModule", "RealModule"]),
```
benrimmington, line 29:
```suggestion
    .target(name: "Numerics", dependencies: ["BigIntModule", "ComplexModule", "RealModule"]),
```

Also:
* This file `Copyright (c) 2019 - 2020`
* Other files `Copyright (c) 2020`

benrimmington: I think the `Numerics` target still needs the `BigIntModule` dependency (see above).

xwu:
@benrimmington Hmm, your review comments just vanished. I've addressed the copyright header situation.

stephentyrone:
Awesome, @xwu. I'll give this a thorough review sometime this week.

Sajjon:
@xwu Excellent performance! Great job! Any plan to include `BigUInt` as well? :)

stephentyrone:
@Sajjon Since a `BigInt`, by definition, can never overflow, there's really no such thing as a `BigUInt` (the distinction between signed and unsigned only makes sense for fixed-width types).

Sajjon:
@stephentyrone Hmm sorry I might misunderstand you, but if I understand it correctly @xwu excellent work so far just declares `BigInt` which is **signed** and thus can allow for negative values. But exactly the same uses cases using Swift Foundation where I might want to model my data with `UInt` instead of `Int` to mark that the value cannot and should not ever be negative, I would like to do the same thing using Big integers, but at compile time, disallow for negative values.

stephentyrone:
@Sajjon Sure, you *can* define such a type, but Swift really discourages using unsigned types just for "a value that can never be negative" (see: sizes being `Int` instead of `UInt`). For fixed-width types, there are other differences that justify having the unsigned types (the arithmetic operations and shifts have different semantics, the set of representable values is different, even if you restrict to all-positive values), but none of those differences hold for non-fixed-width types.

Sajjon:
@stephentyrone Yes I'm quite surprised that `Int` is the default type of Swift Array's size and also Indices... What is the index `-1`? It makes sense as an ephemeral value, like, subtract 1 from an index, but only during that subtraction, which more elegantly is expressed as a throwing subtraction function between `UInt`. So I'm super curious, why did `Int` become the default integer type, instead of `UInt`, at least in the context of Array/Collections? And is that (the context of Arrays/Collections) the only reason why _"Swift really discourages using unsigned types just for...."_ ? 

Instead of using `UInt` for disallowing negative values one can of course wrap it in a custom struct, but since Swift lacks Haskell's `newtype` functionality and since we do not have easy way to delegate conformance of protocols, such as `FixedWidthInteger` (and all other relevant numerical protocols) it is really cumbersome to wrap values if one still wants to use integer functionality on this new type.

stephentyrone:
> So I'm super curious, why did Int become the default integer type, instead of UInt, at least in the context of Array/Collections? And is that (the context of Arrays/Collections) the only reason why "Swift really discourages using unsigned types just for...." ?

Once you decide to eliminate implicit conversions, this decision follows naturally, because otherwise any form of indexing arithmetic becomes super-cumbersome:

- Suppose I don't know the relative order of `i` and `j`, but want to know how far apart they are. If everything is signed, even though they're "always positive", I do `abs(i - j)`. To work with unsigned indices, I would have to use `i < j ? j - i : i - j` or `UInt(abs(Int(i) - Int(j)))`.

- More generally, being able to unconditionally do arithmetic that forms a possibly-out-of-range index, and then check if its in-range before use, is often a much simpler idiom that needing to detect conditions that could lead to forming an out-of-range index before I do it. Consider a really simple convolution with zero extension:
```
for i in a.indices {
  a[i] = (0 ... 5).reduce(into: 0) {
    let j = i - 2 + $1
    if a.indices.contains(j) { $0 += weights[$1]*a[j] }
  }
}
```
If we had unsigned indices, the closure would either look like this:
```
    let j = i + $1
    if j >= 2 || j < a.count + 2 { // can't subtract 2 from `j` until we know it's >= 2.
      $0 += weights[$1]*a[j - 2]
    }
```
or maybe this if you're clever:
```
    let j = i + $1 &- 2 // if this wraps, we'll fail the check below
    if j < a.count {
      $0 += weights[$1]*a[j]
    }
```
These aren't a lot more code, but how they work is significantly less obvious to a new reader.

Sajjon:
@stephentyrone Excellent answer, description and motivation, thank you!

@xwu Would be great if the API of `BigInt` could somewhat reflect the one of `attaswift/BigInt`- it would allow for easy transition - I can help out id you want to? :) Just let me know when the current implementation is somewhat stable (or maybe it is?).

xwu:
> @xwu Would be great if the API of BigInt could somewhat reflect the one of attaswift/BigInt- it would allow for easy transition - I can help out id you want to? :) Just let me know when the current implementation is somewhat stable (or maybe it is?).

All implementations are subject to change, I think. There are only seven APIs here that are introduced beyond what's required by existing protocols, and any differences in their naming as compared to attaswift/BigInt is deliberate, often reflecting what's been formalized in the Swift Evolution process since the time that attaswift/BigInt was created :)

stephentyrone:
> There are only seven APIs here that are introduced beyond what's required by existing protocols

Yeah, the space of integer API is extremely constrained by what's in the stdlib protocols, so there's very little room for variation.

Sajjon:
Ok got it! I saw you've implemented `pow(_ base: Self, _ exponent: Self, modulo modulus: Self) -> Self?` which is great 👍, I've used that a lot in attaswift/BigInt when implementing ECC.

Can't wait to replace (awesome) attaswift/BigInt with just Swift Numerics package everywhere.

Sajjon:
Also, should maybe the [BigInt](https://github.com/apple/swift/blob/master/test/Prototypes/BigInt.swift) prototype in Swift repo under _Prototypes_ be deleted after this PR is merged? Seems like no need to keep it around? Thoughts on that? :)

stephentyrone:
It's used for _testing_ a bunch of Stdlib functionality, so it should stay put for now.

dwaite:
> * Suppose I don't know the relative order of `i` and `j`, but want to know how far apart they are. If everything is signed, even though they're "always positive", I do `abs(i - j)`. To work with unsigned indices, I would have to use `i < j ? j - i : i - j` or `UInt(abs(Int(i) - Int(j)))`.

That and the edge cases - for instance, if the value of the `UInt`s `i` or `j` is greater than `Int.max`, the difference will overflow. Since you may be already be using the 64bit unsigned integer type, dealing with that edge case requires more work. Work that you don't typically deal with in C (not that the compiler there will help you by doing the "right" thing).

swanysteve:
I've been working with the code in this PR.  I've found a problem in the random number generation.

In my case, 
`BigInt.random(in: BigInt(0)...BigInt("12345678901234567890")! )`
works but
`BigInt.random(in: BigInt(0)...BigInt("98765432109876543210")! )`
hangs.

It turns out the first number occupies 64 bits but the second needs 67 bits.

After debugging, in the implementation of
`extension RandomNumberGenerator {
  @usableFromInline
  internal mutating func _next(upperBound: BigInt) -> BigInt {
`
in `RandomNumberGenerator.swift`,
the two cases where results are OR-ed with `mask` should instead be AND-ed with `mask`.  I have seen the `repeat` loop run multiple times, but I no longer see a hang.

I'm very new to Swift and don't yet know how to run the unit tests, but will work on that next.

swanysteve:
While working on a unit test for LCM, I've found a problem with division.

The following test snippet fails
```
    let s = "3182990411991163758463334586237477812181413821081264903177942613807541528844305823312360947978873366530408681494780"
    let t = "16012987498029214696648267754993196043335730812460137832692217701508007910953126091749743662257786118530085930814191"
    let m = (BigInt(s)!, AttaswiftBigInt(s)!)
    let n = (BigInt(t)!, AttaswiftBigInt(t)!)
    XCTAssertEqual((m.0*n.0) / m.0, BigInt((m.1*n.1) / m.1))
```

xwu:
@swanysteve Yikes, I'll have to look into that.

swanysteve:
Thanks @xwu .  My LCM and Pow tests now pass.

I'm still getting a hang in my random test.
```
  func testRandom() {
    let b1 = BigInt(0)
    let t1 = BigInt("12345678901234567890")! //64 bits
    let t2 = BigInt("98765432109876543210")! //67 bits

    func testClosedRanges() {
      let n1 = BigInt.random(in: b1...t1)
      XCTAssert( n1 >= b1 && n1 <= t1);
      let n2 = BigInt.random(in: b1...t2)
      XCTAssert( n2 >= b1 && n2 <= t2);
    }

    func testOpenRanges() {
      let n1 = BigInt.random(in: b1..<t1)
      XCTAssert( n1 >= b1 && n1 < t1);
      let n2 = BigInt.random(in: b1..<t2)
      XCTAssert( n2 >= b1 && n2 < t2);
    }

    testClosedRanges()
    testOpenRanges()
  }
```

xwu:
As you've noted, that's a standard library bug :)

swanysteve:
Sorry, I don't understand.  The bug is in `BigIntModule/RandomNumberGenerator.swift` which is part of this PR, AFAICT.

xwu:
As you can tell, I haven't been focusing on this much. You're absolutely right; will fix that think-o tonight. Thanks for exercising this code, and keep these coming!

1oo7:
What's the hold-up on this PR?

1oo7 (approved):

xwu:
@1oo7 Glad you're enthusiastic about this PR. As you can see, there is some additional work I'd like to do before this gets merged. Karatsuba multiplication is excruciatingly slow, for instance, due to some design decisions that need to be revised. What's more, the PR is waiting on @stephentyrone's comments.

Do feel free to contribute any notes about the implementation, particularly bugs you find. However, in general, it's helpful _not_ to bump PRs, and please avoid using the approval features on GitHub as an upvote.

1oo7:
> @1oo7 Glad you're enthusiastic about this PR. As you can see, there is some additional work I'd like to do before this gets merged. Karatsuba multiplication is excruciatingly slow, for instance, due to some design decisions that need to be revised. What's more, the PR is waiting on @stephentyrone's comments.
> 
> Do feel free to contribute any notes about the implementation, particularly bugs you find. However, in general, it's helpful _not_ to bump PRs, and please avoid using the approval features on GitHub as an upvote.

Thanks Xiaodi. I recognize your name from Swift.org forums were we have had some conversations. Hope to see this progress and also eventually get BigDecimal as well. Thanks for all your hard work.

Sajjon:
Will StaticBigInt change the status of this PR, as in help it progress? https://github.com/apple/swift-evolution/blob/main/proposals/0368-staticbigint.md

xwu:
No doubt!

PR #121, opened 2020-06-08
Define `Complex.zeroTangentVectorInitializer`.

dan-zheng:
The `Differentiable.zeroTangentVectorInitializer` protocol requirement is a closure producing a zero tangent vector, capturing minimal information from `self`.
* [Documentation](https://github.com/apple/swift/blob/tensorflow/docs/DifferentiableProgramming.md#the-differentiable-protocol-1)
* [Default synthesis](https://github.com/apple/swift/blob/tensorflow/docs/DifferentiableProgramming.md#default-synthesis)

Fixes the `Complex: Differentiable` conformance for master toolchains:

```console
$ swift --version
Apple Swift version 5.3-dev (LLVM f516ac602c, Swift 09ea5fd1f7)
Target: x86_64-apple-darwin19.3.0

$ swift build
swift-numerics/Sources/ComplexModule/Differentiable.swift:15:1: error: implementation of 'Differentiable' cannot be automatically synthesized in an extension in a different file to the type
extension Complex: Differentiable
^
swift-numerics/Sources/ComplexModule/Complex.swift:48:15: note: type declared here
public struct Complex<RealType> where RealType: Real {
              ^
```

Necessary after https://github.com/apple/swift/pull/31823, which removed the temporary default implementation of `Differentiable.zeroTangentVectorInitializer`.

### DIFF
--- a/Sources/ComplexModule/Differentiable.swift
+++ b/Sources/ComplexModule/Differentiable.swift
@@ -15,6 +15,11 @@ import _Differentiation
 extension Complex: Differentiable
 where RealType: Differentiable, RealType.TangentVector == RealType {
   public typealias TangentVector = Self
+
+  @inlinable
+  public var zeroTangentVectorInitializer: () -> Self {
+    { Complex.zero }
+  }
 }
 
 extension Complex
--- a/Tests/ComplexTests/DifferentiableTests.swift
+++ b/Tests/ComplexTests/DifferentiableTests.swift
@@ -57,6 +57,10 @@ final class DifferentiableTests: XCTestCase {
     XCTAssertEqual(φDivide(Complex(1, 0)), Complex(0.25, -0.25))
     XCTAssertEqual(φDivide(Complex(0, 1)), Complex(0.25, 0.25))
   }
+
+  func testZeroTangentVectorInitializer() {
+    XCTAssertEqual(Complex<Float>(-5, 5).zeroTangentVector, Complex(0, 0))
+  }
 }
 
 #endif


Sources/ComplexModule/Differentiable.swift
```diff
@@ -15,6 +15,10 @@ import _Differentiation
 extension Complex: Differentiable
 where RealType: Differentiable, RealType.TangentVector == RealType {
   public typealias TangentVector = Self
+
+  public var zeroTangentVectorInitializer: () -> Self {
+    { Complex.zero }
+  }
```
dan-zheng, line 21: I chose not to use a memberwise implementation of `zeroTangentVectorInitializer`:
```swift
  public var zeroTangentVectorInitializer: () -> Self {
    { [realTanInit = real.zeroTangentVectorInitializer,
       imaginaryTanInit = imaginary.zeroTangentVectorInitializer] in
      Complex(realTanInit(), imaginaryTanInit())
    }
  }
```

I don't believe it's necessary for `Complex`.

dan-zheng:
cc @rxwei

rxwei:
Why was this broken? I thought we specifically made the patch that added `zeroTangentVectorInitializer` synthesis be source compatible.

dan-zheng:
`zeroTangentVectorInitializer` synthesis is source-compatible, except that cross-file synthesis is disallowed for all existing derived conformances. I think the diagnostic makes that clear.

We could consider lifting the same-file synthesis restriction just for `Differentiable.zeroTangentVectorInitializer`. I'm not sure if that has negative implications.

Sources/ComplexModule/Differentiable.swift
```diff
@@ -15,6 +15,10 @@ import _Differentiation
 extension Complex: Differentiable
 where RealType: Differentiable, RealType.TangentVector == RealType {
   public typealias TangentVector = Self
+
+  public var zeroTangentVectorInitializer: () -> Self {
```
rxwei, line 19: Perhaps make this `@inlinable`?

dan-zheng:
@stephentyrone: could you please take a look?
This fix helps unblock `apple/swift-numerics` users using Swift master toolchains.

PR #122, opened 2020-06-09
Remove SIMDScalar requirement on numeric type of Quaternion

markuswntr:
This PR removes the SIMD storage on quaternions, replacing it with `(x,y,z,w)` properties and in-turn lifts the requirement of SIMDScalar conformance on the numeric type.

During the discussion on the initial sketch PR it has been brought up that is maybe worth dropping the SIMD storage and the requirement of the SIMDScalar on Quaternion and this PR should help to find the benefits and disadvantages on the API.

### DIFF
--- a/Sources/QuaternionModule/Arithmetic.swift
+++ b/Sources/QuaternionModule/Arithmetic.swift
@@ -15,12 +15,12 @@ import RealModule
 extension Quaternion: AdditiveArithmetic {
   @_transparent
   public static func + (lhs: Quaternion, rhs: Quaternion) -> Quaternion {
-    Quaternion(from: lhs.components + rhs.components)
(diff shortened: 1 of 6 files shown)


stephentyrone:
Yeah, this is enough churn that I think we probably shouldn't bother unless we get an explicit request to support it. Does that seem right to you?

markuswntr:
Yeah, I do agree. I think, we do need to add an explicit 3D vector type over `Real` scalar values before we can go back and think about lifting the SIMDScalar requirement. Some APIs, especially when interacting with the transformations APIs, look much better (more reasonable) on the call site - or we do limit those APIs to quaternions of `SIMDScalar` number types.
Closing the PR - thanks for looking into this!

Issue #3, opened 2019-06-25
Add approximate equality


stephentyrone:
Update [SE-0259](https://forums.swift.org/t/se-0259-approximate-equality-for-floating-point/23627) for inclusion as a module in swift-numerics, and use it for our tests.

stephentyrone:
Marking this issue closed with #123.

PR #123, opened 2020-06-10
Approximate equality comparisons.

stephentyrone:
This is a revised version of SE-0259 with a somewhat simpler API--there's no more special-case handling for zero. It's also more general; it works for any Numeric type whose magnitude conforms to FloatingPoint. This is really nice, since it means that we get Complex and Quaternion support for free.

This new version tracks Julia's [`isapprox`](https://docs.julialang.org/en/v1/base/math/#Base.isapprox) pretty closely; my opinion is that this is the only language with an approximate comparison worth discussing. The divergences from what Julia does are as follows:

- Spelling. Julia is much terser than Swift =)
- Julia defaults to using the Euclidean norm. That's not available in this generic context, and while it's more common, it's actually a less-good choice due to overflow considerations, so I'm defaulting to using the `.magnitude` property.
- When only a relativeTolerance is provided, Julia defaults to absoluteTolerance = 0, but I'm defaulting to absoluteTolerance = relativeTolerance * .leastNormalMagnitude. Perfectly reasonable arguments can be made for both choices; using zero matches the mathematical notion of a pure relative tolerance, but scaling to .leastNormalMagnitude better tracks the actual rounding behavior of floating-point arithmetic.
- Julia falls back on elementwise comparison when `a-b` is non-finite, but that's not possible in Swift, because there's no notion of "element" in the generic context for which this function is defined. I fall back on `a == b` instead. Again, both of these are pretty reasonable choices (neither is great, to be honest, but neither is terrible, either).
- I am not providing the `nans` parameter for now (override behavior so that nan compares approximately equal to nan). This is easy to add later, but it's also an extra knob that I would prefer to keep out of the API if we can. I may very well end up turning around and adding it for testing purposes, however =)

Addresses https://github.com/apple/swift-numerics/issues/3

### DIFF
--- a/Sources/RealModule/ApproximateEquality.swift
+++ b/Sources/RealModule/ApproximateEquality.swift
@@ -67,7 +67,7 @@ extension Numeric where Magnitude: FloatingPoint {
     )
   }
   
-  /// Compares `self` and `other` for approximate equality with a specified tolerances.
+  /// Compares `self` and `other` for approximate equality with specified tolerances.
   ///
   /// `true` if `self` and `other` are equal, or if they are finite and either
   /// ```
--- a/Sources/RealModule/ApproximateEquality.swift
+++ b/Sources/RealModule/ApproximateEquality.swift
@@ -10,16 +10,15 @@
 //===----------------------------------------------------------------------===//
 
 extension Numeric where Magnitude: FloatingPoint {
-  /// Compares `self` and `other` for approximate equality with a specified
-  /// relative tolerance and implicit absolute tolerance.
+  /// Test if `self` and `other` are approximately equal.
   ///
   /// `true` if `self` and `other` are equal, or if they are finite and
   /// ```
-  /// (self - other).magnitude ≤ relativeTolerance * scale
+  /// norm(self - other) <= relativeTolerance * scale
   /// ```
   /// where `scale` is
   /// ```
-  /// max(self.magnitude, other.magnitude, .leastNormalMagnitude)
+  /// max(norm(self), norm(other), .leastNormalMagnitude)
   /// ```
   ///
   /// The default value of `relativeTolerance` is `.ulpOfOne.squareRoot()`,
@@ -28,22 +27,27 @@ extension Numeric where Magnitude: FloatingPoint {
   /// computation being performed, but is not suitable for all use cases.
   ///
   /// Mathematical Properties:
-  /// -
-  /// - `isApproximatelyEqual(to:relativeTolerance:)` is _reflexive_ for
+  /// ------------------------
+  /// 
+  /// - `isApproximatelyEqual(to:relativeTolerance:norm:)` is _reflexive_ for
   ///   non-exceptional values (such as NaN).
   ///
-  /// - `isApproximatelyEqual(to:relativeTolerance:)` is _symmetric_.
+  /// - `isApproximatelyEqual(to:relativeTolerance:norm:)` is _symmetric_.
   ///
-  /// - `isApproximatelyEqual(to:relativeTolerance:)` is __not__ _transitive_.
+  /// - `isApproximatelyEqual(to:relativeTolerance:norm:)` is __not__ _transitive_.
   ///   Because of this, approximately equality is __not an equivalence relation__,
   ///   even when restricted to non-exceptional values.
   ///
   /// - For any point `a`, the set of values that compare approximately equal to `a` is _convex_.
   ///   (Under the assumption that the `.magnitude` property implements a valid norm.)
   ///
-  /// - `isApproximatelyEqual(to:relativeTolerance:)` is _scale invariant_,
+  /// - `isApproximatelyEqual(to:relativeTolerance:norm:)` is _scale invariant_,
   ///   so long as no underflow or overflow has occured, and no exceptional value is produced
   ///   by the scaling.
+  ///
+  /// See Also:
+  /// -------
+  /// - `isApproximatelyEqual(to:absoluteTolerance:[relativeTolerance:norm:])`
   ///
   /// - Parameters:
   ///
@@ -55,32 +59,40 @@ extension Numeric where Magnitude: FloatingPoint {
   ///     This value should be non-negative and less than or equal to 1.
   ///     This constraint on is only checked in debug builds, because a mathematically
   ///     well-defined result exists for any tolerance, even one out of range.
+  ///
+  ///   - norm: The [norm] to use for the comparison.
+  ///     Defaults to `\.magnitude`.
+  ///
+  /// [norm]: https://en.wikipedia.org/wiki/Norm_(mathematics)
   @inlinable @inline(__always)
   public func isApproximatelyEqual(
     to other: Self,
-    relativeTolerance: Magnitude = Magnitude.ulpOfOne.squareRoot()
+    relativeTolerance: Magnitude = Magnitude.ulpOfOne.squareRoot(),
+    norm: (Self) -> Magnitude = \.magnitude
   ) -> Bool {
     return isApproximatelyEqual(
       to: other,
       absoluteTolerance: relativeTolerance * Magnitude.leastNormalMagnitude,
-      relativeTolerance: relativeTolerance
+      relativeTolerance: relativeTolerance,
+      norm: norm
     )
   }
   
-  /// Compares `self` and `other` for approximate equality with specified tolerances.
+  /// Test if `self` and `other` are approximately equal with specified tolerances.
   ///
   /// `true` if `self` and `other` are equal, or if they are finite and either
   /// ```
-  /// (self - other).magnitude ≤ absoluteTolerance
+  /// (self - other).magnitude <= absoluteTolerance
   /// ```
   /// or
   /// ```
-  /// (self - other).magnitude ≤ relativeTolerance * scale
+  /// (self - other).magnitude <= relativeTolerance * scale
   /// ```
   /// where `scale` is `max(self.magnitude, other.magnitude)`.
   ///
   /// Mathematical Properties:
-  /// -
+  /// ------------------------
+  ///
   /// - `isApproximatelyEqual(to:absoluteTolerance:relativeTolerance:)`
   ///   is _reflexive_ for non-exceptional values (such as NaN).
   ///
@@ -94,6 +106,76 @@ extension Numeric where Magnitude: FloatingPoint {
   /// - For any point `a`, the set of values that compare approximately equal to `a` is _convex_.
   ///   (Under the assumption that `norm` implements a valid norm, which cannot be checked
   ///   by this function.)
+  ///
+  /// See Also:
+  /// -------
+  /// - `isApproximatelyEqual(to:[relativeTolerance:])`
+  ///
+  /// - Parameters:
+  ///
+  ///   - other: The value to which `self` is compared.
+  ///
+  ///   - absoluteTolerance: The absolute tolerance to use in the comparison.
+  ///
+  ///     This value should be non-negative and finite.
+  ///     This constraint on is only checked in debug builds, because a mathematically
+  ///     well-defined result exists for any tolerance, even one out of range.
+  ///
+  ///   - relativeTolerance: The relative tolerance to use in the comparison.
+  ///     Defaults to zero.
+  ///
+  ///     This value should be non-negative and less than or equal to 1.
+  ///     This constraint on is only checked in debug builds, because a mathematically
+  ///     well-defined result exists for any tolerance, even one out of range.
+  @inlinable @inline(__always)
+  public func isApproximatelyEqual(
+    to other: Self,
+    absoluteTolerance: Magnitude,
+    relativeTolerance: Magnitude = 0
+  ) -> Bool {
+    self.isApproximatelyEqual(
+      to: other,
+      absoluteTolerance: absoluteTolerance,
+      relativeTolerance: relativeTolerance,
+      norm: \.magnitude
+    )
+  }
+}
+
+extension AdditiveArithmetic {
+  /// Test if `self` and `other` are approximately equal with specified tolerances and norm.
+  ///
+  /// `true` if `self` and `other` are equal, or if they are finite and either
+  /// ```
+  /// norm(self - other) <= absoluteTolerance
+  /// ```
+  /// or
+  /// ```
+  /// norm(self - other) <= relativeTolerance * scale
+  /// ```
+  /// where `scale` is `max(norm(self), norm(other))`.
+  ///
+  /// Mathematical Properties:
+  /// ------------------------
+  ///
+  /// - `isApproximatelyEqual(to:absoluteTolerance:relativeTolerance:norm:)`
+  ///   is _reflexive_ for non-exceptional values (such as NaN).
+  ///
+  /// - `isApproximatelyEqual(to:absoluteTolerance:relativeTolerance:norm:)`
+  ///   is _symmetric_.
+  ///
+  /// - `isApproximatelyEqual(to:absoluteTolerance:relativeTolerance:norm:)`
+  ///   is __not__ _transitive_. Because of this, approximately equality is
+  ///   __not an equivalence relation__, even when restricted to non-exceptional values.
+  ///
+  /// - For any point `a`, the set of values that compare approximately equal to `a` is _convex_.
+  ///   (Under the assumption that `norm` implements a valid norm, which cannot be checked
+  ///   by this function.)
+  ///
+  /// See Also:
+  /// -------
+  /// - `isApproximatelyEqual(to:[relativeTolerance:norm:])`
+  /// - `isApproximatelyEqual(to:absoluteTolerance:[relativeTolerance:])`
   ///
   /// - Parameters:
   ///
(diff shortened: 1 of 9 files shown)


stephentyrone:
https://github.com/apple/swift-numerics/issues/3

stephentyrone:
CC @markuswntr

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,111 @@
...
+  ///   In some cases it is desirable to override the `minimumScale` used in a relative
+  ///   comparison (either by setting it to zero, so that there is no rolloff at zero, or by setting it
+  ///   to a much larger value to account for cancellation). When a `minimumScale` is
+  ///   specified, the computation performed is:
+  ///   ```
+  ///   let scale = max(a.magnitude, b.magnitude, minimumScale)
+  ///   return (a - b).magnitude < t * scale
+  ///   ```
+  ///   The tolerance `t` must be in `.ulpOfOne ..< 1`.  The `minimumScale` must
+  ///   be positive and finite.
```
natecook1000, line 86: Some folks may read "positive" as `> 0` — maybe use "greater than or equal to zero" or "non-negative" here?

stephentyrone: Positive actually means `> 0`, so they'd be correct. I can say "strictly positive" to make it clearer (or spell it as "greater than zero").

kylemacomber:
This is awesome!

That `approximatelyEquals` also covers the zero case is a really nice improvement over [SE-0259](https://github.com/apple/swift-evolution/blob/master/proposals/0259-approximately-equal.md). Though I think I prefer the `isXXXEqual(to:)`-style spelling 😁.

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -9,103 +9,130 @@
...
-    _ other: Self,
-    tolerance: Tolerance<Magnitude> = .relative()
+  public func isApproximatelyEqual(
+    to other: Self,
+    relativeTolerance: Magnitude = 0,
+    absoluteTolerance: Magnitude
   ) -> Bool {
-    // If a and b are actually equal, then they are certainly *almost* equal,
-    // with any allowable tolerance.
+    assert(absoluteTolerance >= 0 && absoluteTolerance.isFinite)
```
xwu, line 130:
```suggestion
    precondition(absoluteTolerance >= 0 && absoluteTolerance.isFinite)
```

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -9,103 +9,130 @@
...
+    assert(relativeTolerance >= 0 && relativeTolerance <= 1)
```
xwu, line 131:
```suggestion
    precondition(relativeTolerance >= 0 && relativeTolerance <= 1)
```

Sources/RealModule/ApproximateEquality.swift
stephentyrone: These are deliberately asserts rather than preconditions; these are not performing a test that's required for safety--the result is perfectly well-defined even when out-of-range tolerances are used, and we don't want to have them in release builds.

stephentyrone:
One thing worth considering for the future is that `Numeric + Magnitude: FloatingPoint` is actually a stronger constraint than we need for this. In the most general setting, we ought to be able to do something like the following:
```
extension AdditiveArithmetic {
  @inlinable @inline(__always)
  public func isApproximatelyEqual<Magnitude: Numeric & Comparable>(
    to other: Self,
    absoluteTolerance: Magnitude,
    relativeTolerance: Magnitude = 0,
    norm: (Self) -> Magnitude
  ) -> Bool {
    assert(
      absoluteTolerance >= 0 && absoluteTolerance.isFinite,
      "absoluteTolerance should be non-negative and finite," +
      "but is \(absoluteTolerance)."
    )
    assert(
      relativeTolerance >= 0 && relativeTolerance <= 1,
      "relativeTolerance should be non-negative and <= 1," +
      "but is \(relativeTolerance)."
    )
    if self == other { return true }
    let delta = norm(self - other)
    let scale = max(norm(self), norm(other))
    let bound = max(absoluteTolerance, scale*relativeTolerance)
    return delta.isFinite && delta <= bound
  }
}
```
The one detail that prevents us from just doing this today is that `isFinite` is not a protocol requirement on `Numeric`. We could define a new protocol that requires it, but it doesn't seem worth it for just this API. Hoisting the property (and some others in the Numerics hierarchy fo the standard library) seems like something worth considering for Swift 6 if we can, because it would enable this type of API.

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// max(self.magnitude, other.magnitude, .leastNormalMagnitude)
+  /// ```
+  ///
+  /// The default value of `relativeTolerance` is `.ulpOfOne.squareRoot()`,
+  /// which corresponds to expecting "about half the digits" in the computed results to be good.
+  /// This is the usual guidance in numerical analysis, if you don't know anything about the
+  /// computation being performed, but is not suitable for all use cases.
+  ///
+  /// Mathematical Properties:
+  /// -
```
kylemacomber, line 31:
```suggestion
  ///
```

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// (self - other).magnitude ≤ absoluteTolerance
+  /// ```
+  /// or
+  /// ```
+  /// (self - other).magnitude ≤ relativeTolerance * scale
+  /// ```
+  /// where `scale` is `max(self.magnitude, other.magnitude)`.
+  ///
+  /// Mathematical Properties:
+  /// -
```
kylemacomber, line 83:
```suggestion
  ///
```

Sources/RealModule/ApproximateEquality.swift
kylemacomber: Unintentional empty bullet?

stephentyrone: It makes Xcode quick help format it as a header:
Screen Shot 2020-07-14 at 3 25 44 PM
vs
Screen Shot 2020-07-14 at 3 25 59 PM

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+    relativeTolerance: Magnitude = Magnitude.ulpOfOne.squareRoot()
+  ) -> Bool {
+    return isApproximatelyEqual(
+      to: other,
+      absoluteTolerance: relativeTolerance * Magnitude.leastNormalMagnitude,
+      relativeTolerance: relativeTolerance
+    )
+  }
+  
+  /// Compares `self` and `other` for approximate equality with a specified tolerances.
```
kylemacomber, line 70:
```suggestion
  /// Compares `self` and `other` for approximate equality with specified tolerances.
```

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+// Copyright (c) 2020 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+extension Numeric where Magnitude: FloatingPoint {
+  /// Compares `self` and `other` for approximate equality with a specified
+  /// relative tolerance and implicit absolute tolerance.
```
kylemacomber, line 14: I assume the majority of folks will opt for this variant? It'd be nice if its summary doc comment was more approachable than the full `isApproximatelyEqual(to:absoluteTolerance:relativeTolerance:norm:)`

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// ```
+  /// or
+  /// ```
+  /// (self - other).magnitude ≤ relativeTolerance * scale
+  /// ```
+  /// where `scale` is `max(self.magnitude, other.magnitude)`.
+  ///
+  /// Mathematical Properties:
+  /// -
+  /// - `isApproximatelyEqual(to:absoluteTolerance:relativeTolerance:)`
```
kylemacomber, line 84: Should these include a `norm` argument label? Or was it deliberately omitted because you don't introduce `norm` until later in the comment?

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+  ///     could use:
+  ///     ```
+  ///     z.isApproximatelyEqual(
+  ///       to: 1,
+  ///       absoluteTolerance: 0.001,
+  ///       norm: \.length
+  ///     )
+  ///     ```
+  ///     (if we used the default, we would be testing if `z` were inside a square region
+  ///     instead.)
```
kylemacomber, line 130: This is a really helpful example! It'd be helpful if there were similar examples of using some of the other possible formulations (e.g. custom `relativeTolerance`, both an `absoluteTolerance` and a `relativeTolerance`).

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+  @inlinable @inline(__always)
+  public func isApproximatelyEqual(
+    to other: Self,
+    absoluteTolerance: Magnitude,
+    relativeTolerance: Magnitude = 0,
+    norm: (Self) -> Magnitude = { $0.magnitude }
+  ) -> Bool {
```
kylemacomber, line 137: I wonder if taken together `isApproximateEqual(to:relativeTolerance:)` and `isApproximateEqual(to:absoluteTolerance:relativeTolerance:norm:)` are a minor violation of the API guideline:

> Prefer to locate parameters with defaults toward the end of the parameter list. Parameters without defaults are usually more essential to the semantics of a method, and provide a stable initial pattern of use where methods are invoked.

because it's not providing "a stable initial pattern"? 🤷‍♂️

Sources/RealModule/ApproximateEquality.swift
stephentyrone: Omitted because it simply becomes too verbose and fewer users should need to use it.

Sources/RealModule/ApproximateEquality.swift
stephentyrone: Yeah, there's a missing intro paragraph here that says something like "use this if you don't know better" (but spelled out in a more welcoming manner).

Sources/RealModule/ApproximateEquality.swift
stephentyrone: I pondered that, but I think it's OK because you can _also_ use `isApproximatelyEqual(to:absoluteTolerance:)`. I don't think there's clear guidance in the API guideline for when a:, b:, and a:b: are all allowed, nor can there be.

Sources/RealModule/ApproximateEquality.swift
stephentyrone: (To be precise about what I mean by "fewer users should need to use it": the `norm` parameter doesn't matter for Real values, which is what I expect to be the overwhelming majority of uses. It only becomes relevant with Complex types or vectors / quaternions / matrices / etc.)

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+    assert(
+      absoluteTolerance >= 0 && absoluteTolerance.isFinite,
+      "absoluteTolerance should be non-negative and finite," +
+      "but is \(absoluteTolerance)."
+    )
+    assert(
+      relativeTolerance >= 0 && relativeTolerance <= 1,
+      "relativeTolerance should be non-negative and <= 1," +
```
stephentyrone, line 145:
```suggestion
      "relativeTolerance should be non-negative and <= 1, " +
```

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -0,0 +1,154 @@
...
+  @inlinable @inline(__always)
+  public func isApproximatelyEqual(
+    to other: Self,
+    absoluteTolerance: Magnitude,
+    relativeTolerance: Magnitude = 0,
+    norm: (Self) -> Magnitude = { $0.magnitude }
+  ) -> Bool {
+    assert(
+      absoluteTolerance >= 0 && absoluteTolerance.isFinite,
+      "absoluteTolerance should be non-negative and finite," +
```
stephentyrone, line 140:
```suggestion
      "absoluteTolerance should be non-negative and finite, " +
```

Sources/RealModule/ApproximateEquality.swift
xwu: Could I suggest maybe extending the underline to make that clear?

xwu:
> In the most general setting, we ought to be able to do something like the following:

I’m curious: What would this behavior look like for integers? And are there other types that would benefit from this more general implementation?

Sources/RealModule/ApproximateEquality.swift
xwu: Makes sense. I think it’d be cool if there were some comment to that effect.

Sources/RealModule/ApproximateEquality.swift
kylemacomber: Is there any good reason Julia `isapprox` lists `rtol` before `atol`?

stephentyrone: I think it's just that they want that to be the privileged option, but they don't have an explicit `isapprox(a,b;rtol)` overload like we do.

kylemacomber: That makes sense. I don't have strong opinion here, but IIUC we could achieve a similar effect by adding another overload (here ignoring `norm`):

```swift
extension Real {
    func isApproximatelyEqual(to other: Self, relativeTolerance: Magnitude = Magnitude.ulpOfOne.squareRoot())
    func isApproximatelyEqual(to other: Self, absoluteTolerance: Magnitude)
    func isApproximatelyEqual(to other: Self, relativeTolerance: Magnitude, absoluteTolerance: Magnitude)
}
```

Is `norm` only useful if you have an `absoluteTolerance`, or should we do something like:

```swift
extension Real {
    func isApproximatelyEqual(to other: Self, relativeTolerance: Magnitude = Magnitude.ulpOfOne.squareRoot(), norm: (Self) -> Magnitude = { $0.magnitude })
    func isApproximatelyEqual(to other: Self, absoluteTolerance: Magnitude, norm: (Self) -> Magnitude = { $0.magnitude })
    func isApproximatelyEqual(to other: Self, relativeTolerance: Magnitude, absoluteTolerance: Magnitude, norm: (Self) -> Magnitude = { $0.magnitude })
}
```

stephentyrone:
Commit message explaining update:

The most general version of the API is now defined on AA as:

  ```
  public func isApproximatelyEqual<Magnitude>(
    to other: Self,
    absoluteTolerance: Magnitude,
    relativeTolerance: Magnitude = 0,
    norm: (Self) -> Magnitude
  ) -> Bool where Magnitude: FloatingPoint {
  ```

Note that AA does not have the magnitude property, so we can't default the norm in this setting. Therefore the following is kept on Numeric:

  ```
  public func isApproximatelyEqual(
    to other: Self,
    absoluteTolerance: Magnitude,
    relativeTolerance: Magnitude = 0
  ) -> Bool {
    self.isApproximatelyEqual(
      to: other,
      absoluteTolerance: absoluteTolerance,
      relativeTolerance: relativeTolerance,
      norm: \.magnitude
    )
  }
  ```

to provide the operation with norm defaulted (note: it's not optional, because then this would call itself).

stephentyrone:
I still need to address a few documentation improvements and write a readme that provides guidance all in one place, but this is pretty close to the final design, I think.

Sources/RealModule/ApproximateEquality.swift
```diff
@@ -128,13 +208,20 @@ extension Numeric where Magnitude: FloatingPoint {
...
-    norm: (Self) -> Magnitude = { $0.magnitude }
-  ) -> Bool {
+    norm: (Self) -> Magnitude
+  ) -> Bool
+  // TODO: constraint should really be weaker than FloatingPoint,
+  // but we need to have `isFinite` for it to work correctly with
+  // floating-point magnitudes in generic contexts, which is the
+  // most common case. The fix for this is to lift the isFinite
+  // requirement to Numeric in the standard library, but that's
+  // source-breaking, so requires an ABI rumspringa.
```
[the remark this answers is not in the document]
stephentyrone: Maybe. That would work for floating point, but there's no semantic guarantee that it actually does something sensible for an unconstrained type.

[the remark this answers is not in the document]
stephentyrone: (There's also another issue in making this too general, which is that the computation of `delta` itself can trap for most integer-derived types; because of that I am loosely leaning towards viewing the Magnitude = FloatingPoint constraint as a good thing here.)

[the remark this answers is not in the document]
xwu: I think it would be reasonable to document some degree of semantic restriction on the 'exceptional' values that `Numeric` can support. I don't know if there'd be any problem with explicitly limiting them to IEEE 754 NaN-or-infinity; can you think of any problems that would arise with that formulation?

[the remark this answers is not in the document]
stephentyrone: I don't see an big problem with it, but we'd have to define "IEEE 754 nan-or-infinity" in some way that makes sense for more general types (e.g. would `Complex`'s single non-finite value fit into that?)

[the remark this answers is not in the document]
stephentyrone: Using `delta - delta == 0` would also explicitly rule-out ever supporting running with the invalid exception unmasked, which is generally not a great plan but some scientific software written in C and Fortran wants to do it for complicated historical reasons (if Swift were to support IEEE 754 exceptions someday, infinity.isFinite should not raise invalid, but `infinity - infinity` should).

stephentyrone:
I'm going to iterate on documenting this a bit, but the actual API surface is what I want it to be in the short term, so I'm going to merge this and do the rest in follow-on PRs.

PR #124, opened 2020-06-14
Refine Hashable and Equatable on Quaternion

markuswntr:
- **Canonical Representations**
Add `canonicalized` and `canonicalizedTransform` properties

- **Equatable**
Update type and header documentation and mention the alternative for transformations in *R³*

- **Hashable**
This PR changes the hashing behavior of `Quaternion`, so that any *finite* quaternion `q` clears the sign bit of its real component before hashing.

- **Transform Equality**
Added a new `transformEquals()` method on `Quaternion` to test for 3D transformation equality.

### DIFF
--- a/Sources/QuaternionModule/Quaternion.swift
+++ b/Sources/QuaternionModule/Quaternion.swift
@@ -365,11 +365,15 @@ extension Quaternion: Hashable {
     return lhs.components == rhs.components
   }
 
-  /// Returns a Boolean value indicating whether the rotation in *R³* of this
-  /// quaternion equals the rotation of `other`.
+  /// Rotation equality comparison
   ///
   /// This method tests for rotation-wise equality in *R³*, where both `q == q`
-  /// but also `q == -q` is `true`.
+  /// but also `q == -q` are `true`.
+  ///
+  /// - Parameters:
+  ///   - other: The value to compare.
+  /// - Returns: Boolean value indicating whether the rotation in *R³* of this
+  /// quaternion equals the rotation of `other`.
   @_transparent
   public func rotationEquals(_ other: Quaternion) -> Bool {
     // Identify all numbers with either component non-finite as a single "point at infinity".
--- a/Sources/QuaternionModule/Quaternion.swift
+++ b/Sources/QuaternionModule/Quaternion.swift
@@ -37,7 +37,7 @@ import RealModule
 /// important to be aware that, when used this way, any quaternion and its
 /// negation represent the same transformation, but they do not compare equal
 /// using `==` because they are not the same quaternion.
-/// You can compare quaternions as 3D transformations using `transformEquals()`.
+/// You can compare quaternions as 3D transformations using `equals(as3DTransform:)`.
 public struct Quaternion<RealType> where RealType: Real & SIMDScalar {
 
   /// The components of the 4-dimensional vector space of the quaternion.
@@ -401,7 +401,7 @@ extension Quaternion: Hashable {
   ///   important to be aware that, when used this way, any quaternion and its
   ///   negation represent the same transformation, but they do not compare
   ///   equal using `==` because they are not the same quaternion. You can
-  ///   compare quaternions as 3D transformations using `transformEquals()`.
+  ///   compare quaternions as 3D transformations using `equals(as3DTransform:)`.
   @_transparent
   public static func == (lhs: Quaternion, rhs: Quaternion) -> Bool {
     // Identify all numbers with either component non-finite as a single "point at infinity".
@@ -413,13 +413,16 @@ extension Quaternion: Hashable {
     return lhs.components == rhs.components
   }
 
-  /// Transformation equality comparison
+  /// Returns a Boolean value indicating whether the 3D transformation of the
+  /// two quaternions are equal.
   ///
-  /// Returns a Boolean value indicating whether the 3D transformations of this
-  /// quaternion equals the 3D transformation of `other`.
+  /// Use this method to test for equality of the 3D transformation properties
+  /// of quaternions; where for any quaternion `q`, its negation represent the
+  /// same 3D transformation; i.e. `q.equals(as3DTransform: q)` as well as
+  /// `q.equals(as3DTransform: -q)` are both `true`.
   ///
   /// - Parameter other: The value to compare.
-  /// - Returns: True if the transformation of this quaternion equals `other`.
+  /// - Returns: True if the 3D transformation of this quaternion equals `other`.
   @_transparent
   public func equals(as3DTransform other: Quaternion) -> Bool {
     // Identify all numbers with either component non-finite as a single "point at infinity".
--- a/Sources/QuaternionModule/Quaternion.swift
+++ b/Sources/QuaternionModule/Quaternion.swift
@@ -421,7 +421,7 @@ extension Quaternion: Hashable {
   /// - Parameter other: The value to compare.
   /// - Returns: True if the transformation of this quaternion equals `other`.
   @_transparent
-  public func transformEquals(_ other: Quaternion) -> Bool {
+  public func equals(as3DTransform other: Quaternion) -> Bool {
     // Identify all numbers with either component non-finite as a single "point at infinity".
     guard isFinite || other.isFinite else { return true }
     // For finite numbers, equality is defined componentwise. Cases where only
--- a/Sources/QuaternionModule/Quaternion.swift
+++ b/Sources/QuaternionModule/Quaternion.swift
@@ -33,9 +33,11 @@ import RealModule
 /// downsides. You can access the Euclidean norm using the `length` property.
 /// See `Complex` type of the swift-numerics package for additional details.
 ///
-/// `==` does not compare rotations in *R³*; it performs a componentwise and
-/// sign sensitive comparison. You can compare the rotations in *R³* of any
-/// two quaternions using `rotationEquals()`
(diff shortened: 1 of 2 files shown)


Sources/QuaternionModule/Quaternion.swift
```diff
@@ -32,6 +32,10 @@ import RealModule
 /// Using the infinity norm avoids this problem entirely without significant
 /// downsides. You can access the Euclidean norm using the `length` property.
 /// See `Complex` type of the swift-numerics package for additional details.
+///
+/// `==` does not compare rotations in *R³*; it performs a componentwise and
+/// sign sensitive comparison. You can compare the rotations in *R³* of any
+/// two quaternions using `rotationEquals()`
```
stephentyrone, line 38: Let's try to make this a little bit less technical. How about something like:

> Quaternions are frequently used to represent 3D transformations. It's important to be aware that, when used this way, any quaternion and its negation represent the same transformation, but they do not compare equal using `==` because they are not the same quaternion. You can compare quaternions as 3D transformations using `transformEquals()`.

(Note, "transformEquals" because these aren't pure rotations. Does that seem ok to you?) I think we should also add a "canonicalize transform" operation that selects the representative with sign bit of `x` clear, maps all zeros to (+0, +0, +0, +0), and maps all infinites to a canonical representative (r = .infinity, (x,y,z) zero?).

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -358,8 +393,18 @@ extension Quaternion: Hashable {
     // representation. The correct behavior for zero falls out for free from
     // the hash behavior of floating-point, but we need to use a
     // representative member for any non-finite values.
+    // For any values not zero and infinity we use a canonical form, so
+    // that q and -q hash to the same value. This allows people who are using
+    // quaternions as rotations to get the expected semantics out of collections
+    // (while unfortunately producing some collisions for people who are not,
+    // but not in too catastrophic of a fashion).
```
stephentyrone, line 400: This is probably overkill. I think I would just canonicalize to `r.sign == .plus` and hash.

Sources/QuaternionModule/Quaternion.swift
markuswntr: Thats definitely a nicer documentation on the rotation/transform equality behavior, thank you very much!
I do agree that `transformEquals` is probably a better suited method name for a generic quaternion - I'll change it.
Does a property called `canonicalized` – like on `Complex` – seem ok to you for the "canonicalize transform" operation or is it misleading since it (might) also flip the sign of `real`?

Sources/QuaternionModule/Quaternion.swift
markuswntr: I am sorry, I misinterpreted you comment on the sketch PR.
I guess I'll change it to hash the "canonicalized transform" representation then.

I do see/agree now, that the collisions for people who are not using quaternions for rotations are not too catastrophic :)

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -248,6 +254,57 @@ extension Quaternion {
...
+  /// components are canonicalized in the result.
+  ///
+  /// See also:
+  /// -
+  /// - `.canonicalized`
+  @_transparent
+  public var canonicalizedTransform: Self {
+    var canonical = canonicalized
+    if canonical.real.sign == .plus { return canonical }
+    // Clear the signbit of real even for -0
```
stephentyrone, line 304: I think this might be clearer if we just write `return -canonical`.

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -339,6 +396,17 @@ extension Quaternion where RealType: BinaryFloatingPoint {
 
 // MARK: - Conformance to Hashable and Equatable
 extension Quaternion: Hashable {
+  /// Returns a Boolean value indicating whether two values are equal.
+  ///
+  /// Equality is the inverse of inequality. For any values *a* and *b*,
+  /// `a == b` implies that `a != b` is `false`.
```
stephentyrone, line 402: This is true for anything conforming to `Equatable`, so I don't think that we need to spell it out here; is there a specific motivation for adding it?

Tests/QuaternionTests/PropertyTests.swift
```diff
@@ -100,13 +100,80 @@ final class PropertyTests: XCTestCase {
       XCTAssertEqual(infs[0], i)
       XCTAssertEqual(infs[0].hashValue, i.hashValue)
     }
+    // Validate that all *normal* values hash their absolute components, so
+    // that rotations in *R³* of `q` and `-q` will hash to same value.
+    let pairs: [(lhs: Quaternion<T>, rhs: Quaternion<T>)] = [
+      (
+        Quaternion<T>(real:  .pi, imaginary:  .pi,  .pi,  .pi),
+        Quaternion<T>(real: -.pi, imaginary:  .pi,  .pi,  .pi)
```
stephentyrone, line 108: These should not hash to the same value; the pairs should be `-pi, pi, pi, pi` and `pi, -pi, -pi, -pi`, right?

Sources/QuaternionModule/Quaternion.swift
markuswntr: Changed it, thank you!

Sources/QuaternionModule/Quaternion.swift
markuswntr: No specific reason. I thought it might be nice to have same comments as `==` from `Equatable`. I removed it.

Tests/QuaternionTests/PropertyTests.swift
markuswntr: No, that's not correct - my bad!
The `canonicalTransform` implementation had a bug and these tests are wrong too.

Bug fixed and tests adjusted - thank you!

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -248,6 +254,55 @@ extension Quaternion {
...
+  /// -
+  /// - `.canonicalizedTransform`
+  @_transparent
+  public var canonicalized: Self {
+    guard !isZero else { return .zero }
+    guard isFinite else { return .infinity }
+    return self.multiplied(by: 1)
+  }
+
+  /// A "canonical transformation" representation of the value.
```
stephentyrone, line 284: I'd like to make this more precise. If I wasn't concerned about jargon, I would say "the canonical representative of the equivalence class of quaternions representing the same 3d transformation."

Sources/QuaternionModule/Quaternion.swift
```diff
@@ -350,6 +413,24 @@ extension Quaternion: Hashable {
...
 
+  /// Transformation equality comparison
+  ///
+  /// Returns a Boolean value indicating whether the 3D transformations of this
+  /// quaternion equals the 3D transformation of `other`.
+  ///
+  /// - Parameter other: The value to compare.
+  /// - Returns: True if the transformation of this quaternion equals `other`.
+  @_transparent
+  public func transformEquals(_ other: Quaternion) -> Bool {
```
stephentyrone, line 424: How do we feel about how this reads at a use site?
```
if q.transformEquals(p) { /* ... */ }
```
I don't love that "transform" may read as a verb if you don't know what you're looking at. What if we used something like this instead:
```
if q.equals(as3DTransform: p) { /* ... */ }
```
I'm not totally set on this either way, but let's give it some thought.

markuswntr: > How do we feel about how this reads at a use site?
> 
> ```
> if q.transformEquals(p) { /* ... */ }
> ```

Yeah, you are right. This can easily be misread. We could maybe use `transformationEquals` to make it more explicit.

> ```
> if q.equals(as3DTransform: p) { /* ... */ }
> ```

I think, I do prefer this spelling over `transformEquals`/`transformationEquals`. If we want to (somewhat) reuse existing naming patterns, we could utilize `append(contentsOf:)` and `insert(contentsOf:at:)` on `Array` and use:

```swift
if q.equals(transformationOf: p) { /* ... */ }
```

Though, I think it is a little less obvious and does not mention 3D explicitly.

stephentyrone: Let's think about this over the weekend and see if we can make a provisional decision on Monday.

Sources/QuaternionModule/Quaternion.swift
markuswntr: "The canonical representative of quaternions with non-negative real components representing the same 3D transformation."
Does that make sense? :)

Sources/QuaternionModule/Quaternion.swift
stephentyrone: Any further thoughts on this? If not, let's go with `equals(as3DTransform:)` for now?

NevinBR: Perhaps `q.isSame3DTransform(as: p)` or some such?

markuswntr: I changed it to `equals(as3DTransform:)` for now. I think, it reads a little nicer at the call site.

stephentyrone:
@markuswntr is this ready to merge from your perspective, or do you want to iterate on it more first?

markuswntr:
Hi @stephentyrone, as you mentioned previously, the `canonicalizedTransform` property needs a more precise header comment. Other than that I think it is ready to merge.
If it is okay with you, I will provide an updated documentation later - than you can merge it as is.

PR #125, opened 2020-06-18
Add transformation operations to Quaternion

markuswntr:
This PR adds an API for working with other representations of transformations, such as [*Angle-Axis*][angle_axis_wiki], [*Polar*][polar_wiki] and [*Rotation Vector*][rotation_vector_wiki]. The API provides operations to convert from these representations to `Quaternion` and vice versa.

Additionally, the API provides a method to directly rotate an arbitrary vector by a quaternion and thus avoids the calculation of an intermediate representation to any other form in the process.

[angle_axis_wiki]: https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Recovering_the_axis-angle_representation
[polar_wiki]: https://en.wikipedia.org/wiki/Polar_decomposition#Quaternion_polar_decomposition
[rotation_vector_wiki]: https://en.wikipedia.org/wiki/Axis–angle_representation#Rotation_vector

### DIFF
--- a/Sources/QuaternionModule/Transformation.swift
+++ b/Sources/QuaternionModule/Transformation.swift
@@ -121,8 +121,8 @@ extension Quaternion {
 
   /// The [polar decomposition][wiki].
   ///
-  /// Returns the length of this quaternion, half rotation angle in radians of
-  /// *[0, π]* range and the rotation axis as SIMD3 vector of unit length.
+  /// Returns the length of this quaternion, phase in radians of range *[0, π]*
+  /// and the rotation axis as SIMD3 vector of unit length.
   ///
   /// Edge cases:
   /// -
@@ -144,7 +144,7 @@ extension Quaternion {
   /// - `init(rotation:)`
   ///
   /// [wiki]: https://en.wikipedia.org/wiki/Polar_decomposition#Quaternion_polar_decomposition
-  public var polar: (length: RealType, halfAngle: RealType, axis: SIMD3<RealType>) {
+  public var polar: (length: RealType, phase: RealType, axis: SIMD3<RealType>) {
     (length, halfAngle, axis)
   }
 
@@ -257,11 +257,11 @@ extension Quaternion {
 
   /// Creates a quaternion specified with [polar coordinates][wiki].
   ///
-  /// This initializer reads given `length`, `halfAngle` and `axis` values and
+  /// This initializer reads given `length`, `phase` and `axis` values and
   /// creates a quaternion of equal rotation properties and specified *length*
   /// using the following equation:
   ///
-  ///     Q = (cos(halfAngle), axis * sin(halfAngle)) * length
+  ///     Q = (cos(phase), axis * sin(phase)) * length
   ///
   /// Given `axis` gets normalized if it is not of unit length.
   ///
@@ -293,11 +293,11 @@ extension Quaternion {
   ///
   /// [wiki]: https://en.wikipedia.org/wiki/Polar_decomposition#Quaternion_polar_decomposition
   @inlinable
-  public init(length: RealType, halfAngle: RealType, axis: SIMD3<RealType>) {
+  public init(length: RealType, phase: RealType, axis: SIMD3<RealType>) {
     let axisLength: RealType = .sqrt(axis.lengthSquared)
-    if halfAngle.isFinite && axisLength.isNormal {
+    if phase.isFinite && axisLength.isNormal {
       self = Quaternion(
-        halfAngle: halfAngle,
+        halfAngle: phase,
         unitAxis: axis/axisLength
       ).multiplied(by: length)
     } else {
--- a/Sources/QuaternionModule/Transformation.swift
+++ b/Sources/QuaternionModule/Transformation.swift
@@ -443,7 +443,7 @@ extension Quaternion {
   ///   - unitAxis: The rotation axis of unit length
   @usableFromInline @inline(__always)
   internal init(halfAngle: RealType, unitAxis: SIMD3<RealType>) {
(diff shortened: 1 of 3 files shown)


Sources/QuaternionModule/Transformation.swift
```diff
@@ -0,0 +1,415 @@
...
+  /// - `.polar`
+  /// - `init(rotation:)`
+  /// - `init(length:angle:axis)`
+  ///
+  /// - Parameter angle: The rotation angle about the rotation axis in radians
+  /// - Parameter axis: The rotation axis
+  ///
+  /// [wiki]: https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Recovering_the_axis-angle_representation
+  @inlinable
+  public init(angle: RealType, axis: SIMD3<RealType>) {
```
stephentyrone, line 193: Should this take an explicit `length` parameter (with default value 1), so that it's dual to `polar`?

Sources/QuaternionModule/Transformation.swift
```diff
@@ -0,0 +1,415 @@
...
+      )
+      self = Quaternion(length)
+    }
+  }
+
+  /// Rotates given vector by this quaternion.
+  ///
+  /// As quaternions can be used to represent three-dimensional rotations, it is
+  /// is possible to also rotate a three-dimensional vector by a quaternion. The
+  /// rotation of an arbitrary vector by a quaternion is known as an action.
```
stephentyrone, line 316: Let's spell out what the action is a little bit more clearly here. `qvq⁻¹` should appear somewhere in this description =)

Sources/QuaternionModule/Transformation.swift
markuswntr: > * Should this take an explicit `length` parameter (with default value 1), so that it's dual to `polar`?

Thats probably a great idea. Also, I was a bit worried that people might mistakenly use the polar initializer for a scaled Quaternion from an Angle/Axis representation, when the really should not. This would help separate them both clearly and hopefully guide people to the correct initializer. I'll add it, thank you!

> * Instead of normalizing, we could require that `axis` is normal (with a debug assert check?)

I think, at some point, it would be nice to add a construct that guarantees unity of an underlying AlgebraicField type (something like `struct Unit<T: AlgebraicField> { var normalized: T? }`) and only those are assumed to have unit length, while others are not. However, `axis` is arguably always of unit length (or should be), and as far as I can tell, most other libraries just assume it is of unit length, so I am a bit back and forth on this. If we go with a debug assert check, might I ask, what would be a reasonable tolerance for the length of the axis?

Sources/QuaternionModule/Transformation.swift
markuswntr: I updated the comments and wrote down the action more clearly.

Sources/QuaternionModule/Transformation.swift
markuswntr: I implemented both of the suggested changes (Thanks!).
The initializer now features an additional length parameter (defaults to 1, if omitted) and the axis is assumed to be of unit length – instead of normalizing it – or it will raise an assertion failure.

However, there are now two initializer available that are almost equal, namely:

- `init(length: RealType = 1, angle: RealType, axis: SIMD3<RealType>)` for Angle-Axis
- `init(length: RealType, phase: RealType, axis: SIMD3<RealType>)` for Polar decomposition

and I am uncertain if this acceptable/fine.

stephentyrone (approved):

Sources/QuaternionModule/Transformation.swift
```diff
@@ -0,0 +1,487 @@
...
+    let p1 = vector * (real*real - imaginary.lengthSquared)
+    let p2 = 2 * imaginary * imaginary.dot(vector)
+    let p3 = 2 * real * imaginary.cross(vector)
+    let rotatedVector = p1 + p2 + p3
+
+    // If the rotation computes without over/underflow, everything is fine
+    // and the result is correct. If not, we have to do the computation
+    // carefully and first unscale the vector, rotate it again and then
+    // rescale the vector
+    if rotatedVector.isNormal { return rotatedVector }
```
madbat, line 398: Hi! Is this check correct?
I'm using `quaternion.act(on: (1, 0, 0))`, thus `rotatedVector` is `(1, 0, 0)`, which by this module definition of `isNormal` always fails, leading to infinite recursion.

markuswntr: Hi @madbat , thanks so much for reporting the issue!
I am happy to look into possible issues concerning the implementation. Just a heads-up, it might take a few days.
In the meantime, would you mind sharing more insights into how the quaternion is constructed, or more specifically what are the values of the quaternion acting on the vector? Is the quaternion of unit length and/or is the quaternion itself "normal"?

madbat: Thanks for the quick reply @markuswntr.

Here's my code which leads to the issue:
```
let q = Quaternion(angle: Double.pi/2, axis: SIMD3(0, 1, 0))
let transformed = q.act(on: SIMD3(1, 0, 0))
```

markuswntr: Hi @madbat, thanks for sharing the code.
You were correct. The implementation did not properly account for zero values.
I attempt to provide a more elegant implementation before raising a PR, but this may take a while.
For now I provided a quick fix to the issue at: https://github.com/markuswntr/swift-numerics/tree/quaternion/patch-2
It is not heavily tested, but solves your specific issue and should not cause any side-effects.
Feel free to raise further issues with the current/new implementation and/or suggestions!

madbat: Awesome, thank you @markuswntr!

PR #128, opened 2020-07-17
[AutoDiff] Add JVPs for `Complex`

efremale:
Add JVPs to https://github.com/apple/swift-numerics/blob/master/Sources/ComplexModule/Differentiable.swift

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,7 +27,13 @@ let package = Package(
     .target(name: "_NumericsShims", dependencies: []),
     .target(name: "RealModule", dependencies: ["_NumericsShims"]),
     
-    .testTarget(name: "ComplexTests", dependencies: ["ComplexModule", "_NumericsShims"]),
+    .testTarget(
+      name: "ComplexTests",
+      dependencies: ["ComplexModule", "_NumericsShims"],
+      swiftSettings: [
+        .unsafeFlags(["-Xfrontend", "-enable-experimental-forward-mode-differentiation"])
+      ]
+    ),
     .testTarget(name: "RealTests", dependencies: ["RealModule"]),
   ]
 )
--- a/Sources/ComplexModule/Differentiable.swift
+++ b/Sources/ComplexModule/Differentiable.swift
@@ -27,15 +27,21 @@ where RealType: Differentiable, RealType.TangentVector == RealType {
 
   @derivative(of: real)
   @usableFromInline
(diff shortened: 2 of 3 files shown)


efremale:
cc @dan-zheng, @rxwei

Sources/ComplexModule/Differentiable.swift
```diff
@@ -27,15 +27,21 @@ where RealType: Differentiable, RealType.TangentVector == RealType {
 
   @derivative(of: real)
   @usableFromInline
-  func _derivativeReal() -> (value: RealType, pullback: (RealType) -> Complex) {
+  func _vjpDerivativeReal() -> (value: RealType, pullback: (RealType) -> Complex) {
```
dan-zheng, line 30: For clarity, I think it's better to omit "derivative" from these function names and just use "jvp" or "vjp":
```suggestion
  func _vjpReal() -> (value: RealType, pullback: (RealType) -> Complex) {
```

dan-zheng: FYI about a future direction: since `@derivative` functions are never called directly, it would be nice to declare them as "anonymous functions" ([SR-12003](https://bugs.swift.org/browse/SR-12003)), if they're ever added to Swift. That avoids having to pick a name for these functions and prevents them from showing up in code completion.

Package.swift
```diff
@@ -27,7 +27,13 @@ let package = Package(
     .target(name: "_NumericsShims", dependencies: []),
     .target(name: "RealModule", dependencies: ["_NumericsShims"]),
     
-    .testTarget(name: "ComplexTests", dependencies: ["ComplexModule", "_NumericsShims"]),
+    .testTarget(
+      name: "ComplexTests",
+      dependencies: ["ComplexModule", "_NumericsShims"],
+      swiftSettings: [
+        .unsafeFlags(["-Xfrontend", "-enable-experimental-forward-mode-differentiation"])
```
stephentyrone, line 34: What compiler frontends support this, and what are the implications for projects that include Swift Numerics if we add an unsafe setting?

stephentyrone: Can you rebase this and put guards around anything that needs upstream swift changes to require swift > 5.3?

stephentyrone:
Can you rebase this and put guards around anything that needs upstream swift changes to require swift > 5.3?

PR #134, opened 2020-07-27
Remove anything that depends on differentiable programming from master

stephentyrone:
This turns out to not be fully upstreamed such that it doesn't work correctly in 5.2.x on Linux. We'll keep it around on the swift-5.3 branch (because the Swift toolchain there has support).

Looking ahead, I'd like to do any feature work that depends on unreleased Swift toolchain support on a "future" branch.


stephentyrone:
@buttaface I would prefer to use the following approach instead: https://github.com/apple/swift-numerics/pull/136

Can you check to make sure that works for your system? Thanks in advance.

stephentyrone:
Closing in favor of #136

PR #137, opened 2020-08-05
build: add new source file to `RealModule`

compnerd:
`Float16+Real.swift` was added to the source tree, update the build
rules.


stephentyrone:
Thanks!

Issue #2, opened 2019-06-25
Conform `Complex` to `ElementaryFunctions`


stephentyrone:
All of the usual `ElementaryFunctions` operations should be defined for `Complex` types. Implementation details may be found in Kahan's "Branch Cuts for Complex Elementary Functions; or Much Ado About Nothing's Sign Bit".

We can simplify the implementation somewhat for Swift complex numbers, because:
(a) we do not need to worry about fenv flags
(b) we do not need to bother with sign of zero/inf, because we're not trying to be precise with those cases (they do not pull their weight, complexity-wise).

NevinBR:
@stephentyrone, is there something holding this up?

stephentyrone:
Finite number of hours in a day, plus some higher-priority projects to work on.

stephentyrone:
Adding a concrete schedule here: I'll make some time to wrap this up during (US) thanksgiving week, so it should be available before the end of the month.

NevinBR:
> I would suggest phrasing more like "Is there anything I can do to help with this feature?" in general.

Thanks, for the feedback, that’s good advice.

I intended the question to be about the mathematical / computational side, as in, do we need to make any difficult decisions regarding tradeoffs when selecting specific formulas / algorithms for these functions, or would it be acceptable to drop in “textbook” implementations and then potentially refine them later for, eg. numerical stability.

stephentyrone:
Ah. These are mostly fairly-straightforward.

The only real issue is that there are multiple "textbook" formulas for all of these, and we do need to take some care about:
- ensuring that the branch cuts fall in the "right" place (Kahan's 1986 paper "Branch Cuts for Complex Elementary Functions" has good choices for all of these).
- avoiding spurious overflow/underflow (this is a little bit trickier in a generic context than what Kahan deals with in his paper, but definitely isn't new research or anything).

I'm not going after last-bit rounding (or really even close to it) at this point. I *will* be leaving notes about what would need to change to enable it. The biggest pile of work is actually writing test cases, not the implementations themselves.

stephentyrone:
This has been done for a while now.

PR #142, opened 2020-08-14
Make Complex conform to ElementaryFunction

dankogai:
This PR addresses https://github.com/apple/swift-numerics/issues/2 which appears to be long overdue.

> The biggest pile of work is actually writing test cases, not the implementations themselves.

The best way to do so is to borrow what is already working.  We already have a working complex math in C99 so I borrowed it.  the sole purpose of `Sources/_CComplex` is to offer the functions that `<complex.h>` offers.  And `Tests/ComplexTests/ElementaryFunctionsTest.swift` does the test.

The implementation at `Sources/ComplexModule/Complex+ElementaryFunctions.swift` is pretty straightforward.  Which was by far the easiest part of this PR.


xwu:
Hey @dankogai! It's nice to see forward progress on this. I have a version of this working in native Swift here: https://github.com/xwu/NumericAnnex/blob/master/Sources/Complex.swift

It needs to be adapted to how this library handles the sign of zero/infinity, but otherwise should be a quick cut-and-paste (you or anyone else has my permission). That way, we avoid any further shims and provide consistent results across platforms.

What's missing are robust test cases, as you quote, and I wonder if you wouldn't mind taking a crack at that here.

HaydenMcCabe:
Hi,

I haven't had time to work on this branch for a bit, but I do have some test functions. They're based on having the shims, as they compare the native Swift implementations against those provided by LLVM's builtins. There are also some utility functions that generate random floating point numbers from the entire range available for the type, rather than the Swift functions on the types that generate random numbers of the same size (i.e., with similar exponents.) I could adapt those test functions to work on known testing values, or the shims could be included as a dependency for only the test cases.

There are also test cases that rate the performance of the native Swift implementations against those provided by the builtins, but those are mostly to benchmark the performance I was getting during development and don't need to be in the final code. Several of the implementations I did were based on Taylor series, so I wanted to see if they were converging in a reasonable amount of time.

You can see the code on this branch: https://github.com/HaydenMcCabe/swift-numerics/tree/ComplexElementaryFunctions/Tests/ComplexTests

I'm going to be out today, but I can take a look at it later.

dankogai:
Hi,

One request. Avoid Taylor series within `Complex` with fixed number of iterations(as [asin() in your repo]). That would make it very hard, if not impossible, to use arbitrary precision `FloatingPoint` as its elements.

[asin() in your repo]: https://github.com/HaydenMcCabe/swift-numerics/blob/ComplexElementaryFunctions/Sources/ComplexModule/Complex%2BElementaryFunctions.swift

As a matter of fact I am developing one, [swift-bignum].  And I made it comform to `Real` so it can be used as an element of `Complex` of `ComplexModule`.

[swift-bignum]: https://github.com/dankogai/swift-bignum

As you can see below the precision is adjustable because it internally adjusts the number of iterations needed to achieve the precision that the user ordered.  Let its elements take care of the calculations. `Complex` is generic so keep it as generic as possible.

```shell
% swift run --repl
Launching Swift REPL with arguments: -I/Users/dankogai/github/swift-bignum/.build/x86_64-apple-macosx/debug -L/Users/dankogai/github/swift-bignum/.build/x86_64-apple-macosx/debug -lBigNum__REPL -I/Users/dankogai/github/swift-bignum/.build/checkouts/swift-numerics/Sources/_CComplex/include -I/Users/dankogai/github/swift-bignum/.build/checkouts/swift-numerics/Sources/_NumericsShims/include
Welcome to Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53).
Type :help for assistance.
  1> import Numerics 
  2> import BigNum 
  3> Complex.log(Complex(-1.0)) 
$R0: ComplexModule.Complex<Double> = {
  x = 0
  y = 3.1415926535897931
}
  4> Complex.log(Complex(BigFloat(-1.0)))
$R1: ComplexModule.Complex<BigNum.BigFloat> = {
  x = {
    scale = 0
    mantissa = {
      magnitude = {
        kind = inline {
          inline = {
            0 = 0
            1 = 0
          }
        }
        storage = 0 values
      }
      sign = plus
    }
  }
  y = {
    scale = -62
    mantissa = {
      magnitude = {
        kind = inline {
          inline = {
            0 = 14488038916154245685
            1 = 0
          }
        }
        storage = 0 values
      }
      sign = plus
    }
  }
}
  5> Complex.log(Complex(BigFloat(-1.0))).description
$R2: String = "(0.0, 3.14159265358979323851)"
  6> BigFloat.precision
$R3: Int = 64
  7> BigFloat.precision = 128
  8> Complex.log(Complex(BigFloat(-1.0))).description 
$R4: String = "(0.0, 3.141592653589793238462643383279502884195)"
  9>  
```

NevinBR:
I observe this PR has no tests for `expMinusOne(_:)` and `log(onePlus:)`.

Also, and I haven’t actually checked, but at a glance the proposed implementations of those functions do not appear to preserve full precision for inputs near zero.

stephentyrone:
> The best way to do so is to borrow what is already working. We already have a working complex math in C99 so I borrowed it. the sole purpose of Sources/_CComplex is to offer the functions that <complex.h> offers.

This is unfortunately not viable on Windows, which has a rather primitive C "standard library". That's not to say we can't take it, but we'll need to have full implementations of all of these anyway, and we can certainly do better than most C libraries do (the <complex.h> functions have never seen any love on most platforms).

NevinBR:
I see you’ve updated the implementations of `expMinusOne` and `log(onePlus:)`.

Now, I’m not trying to be “that guy”, and I appreciate that you gave credit to the place you found those implementations, but they were actually written by me [here](https://forums.swift.org/t/elementaryfunctions-compliance-for-complex/37903/3) on the Swift forums, 2 weeks before Hayden added them to his repo in [this commit](https://github.com/HaydenMcCabe/swift-numerics/commit/432c1e0d9cee5aa68ad7ce147f3645534aaeee84).

And at the very least, the `log(onePlus:)` implementation should check for spurious overflow when calculating `a`, something like this:

```
extension Complex {
  static func log(onePlus z: Self) -> Self {
    let a = z.lengthSquared + 2 * z.x
    if !a.isFinite { return log(z + Self(1)) }
    let u = RealType.log(onePlus: a) / 2
    let v = RealType.atan2(y: z.y, x: 1 + z.x)
    return Self(u, v)
  }
}
```

Also, both of these implementations are non-obvious enough that they probably deserve some comments explaining the derivation and error analysis.

stephentyrone:
Hi @dankogai!

When I said

> The biggest pile of work is actually writing test cases, not the implementations themselves.

what I was referring to was choosing _input_ values, not having a reference to test against. Leveraging the C complex library (where available) gets a reference, but doesn't help with choosing the right inputs to use for testing.

E.g. the proposed `exp` implementation spuriously overflows for an input like (710, .pi/4)--the result should be finite, but isn't. `cosh` and `sinh` have the same problem. There needs to be test cases that exercises that.

`tanh` as implemented produces infinity for (1000,0), when the result should be 1.0.

`log` uses `.polar`, which overflows for values like (.greatestFiniteMagnitude, .greatestFiniteMagnitude), which would result in incorrect answers when there's a well defined solution.

`sqrt` suffers from catastrophic cancellation (when the imaginary part of the input is small) as well as spurious overflow.

These are the problems that I notice from visual inspection, _without_ a robust test suite in place. All of these are fairly straightforward to correct, fortunately, but I hope this illustrates why I've been holding landing these back while I chip away on writing some tests.

What I'm going to do is start landing my WIP on this, so that other people can help chip away at it one function at a time, rather than facing the burden of tackling the whole problem at once. I'll put up a PR with that tomorrow. I hope that you and others might contribute some smaller PRs that we can review carefully and ensure we have appropriate test coverage--I'd like to aim for one function + anything trivially defined in terms of it per PR.

PR #144, opened 2020-08-24
Factor test support out into a module

stephentyrone:
So that both real and complex and other modules can use it.


stephentyrone:
CC @compnerd for CMake changes

Sources/_TestSupport/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+#[[
+This source file is part of the Swift Numerics open source project
+
+Copyright (c) 2019-2020 Apple Inc. and the Swift Numerics project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+add_library(TestSupport
```
compnerd, line 10: I think that this should be named `_TestSupport` rather than `TestSupport`.  Otherwise, this is generates a module with the name `TestSupport.dll`/`libTestSupport.dylib`/`libTestSupport.so` and is imported as `TestSupport`.  If you want just the import name to be `_TestSupport` the `Swift_MODULE_NAME` property will let you alter just that.

stephentyrone: Should be resolved now. Just a typo.

PR #145, opened 2020-08-24
Added mulAdd and cosMinusOne operations for Real.

stephentyrone:
Defines two new operations for types conforming to Real:

```
/// cos(x) - 1, computed in such a way as to maintain accuracy for small x.
static func cosMinusOne(_ x: Self) -> Self

/// a*b + c, computed _either_ with an FMA or with separate multiply and add.
static func _mulAdd(_ a: Self, _ b: Self, _ c: Self) -> Self
```

These are _mostly_ implementation hooks for Complex, but are generally useful, so I've made them public. Experimental, subject to renaming, especially mulAdd. The first is "trivial" to define using the half-angle formula, but it's a sufficiently useful implementation hook that it's worth having.

Naming of `cosMinusOne` is clear enough; `_mulAdd` is more of a niche feature, and the naming isn't obvious, so I'm underscoring it for now.

### DIFF
--- a/Sources/RealModule/Double+Real.swift
+++ b/Sources/RealModule/Double+Real.swift
@@ -218,4 +218,9 @@ extension Double: Real {
     return libm_lgamma(x, &dontCare)
   }
   #endif
+  
+  @_transparent
+  public static func _mulAdd(_ a: Double, _ b: Double, _ c: Double) -> Double {
+    _numerics_muladd(a, b, c)
+  }
 }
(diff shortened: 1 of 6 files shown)


stephentyrone:
CC @NevinBR who has asked about functionality like `_mulAdd` in the past. It's not exactly what you've asked for, but it's sufficiently useful that you may want to borrow and adapt it.

PR #150, opened 2020-09-02
Refactor README.md and include Approximate Equality as part of the RealModule.

markuswntr:
Approximate Equality has recently been merged into master. This PR updates the README.md to reflect the changes.
While it is part of the `RealModule` it really is an extension on `Numeric`, so I decided to list it below the `RealModule` but separately.


8bitmp3:
Hey @markuswntr Thank for the PR. Big fan of Swift Numerics.

Take a look at https://github.com/markuswntr/swift-numerics/pull/1 - I've made a few improvement suggestions to the entire README based on tech writing experience - hope you like them and we can bundle the changes in this PR: https://github.com/apple/swift-numerics/pull/150:

Summary of suggestions:

- Small change: "may migrate in the second" -> "may migrate into the second"
- Refactor: "new uses are discovered" (passive voice) -> "more users start using Swift Numerics." (active voice)
- Reduce the use of "simply" if possible (it's not necessary but is recommended)
- Refactor + replace the idiom ("to pull in") for internalization/inclusiveness:

```diff
- Swift Numerics modules are fine-grained; if you need support for Complex
+ Swift Numerics modules are fine-grained. For example, if you need support for
+ Complex numbers ...
- ... import ComplexModule¹ without pulling in everything else in the library:
+ ... import ComplexModule¹ as a standalone module:
...
- as well.
...
[Example in code blocks]
```

- Refactor:

```diff
- // All swift-numerics API is now available
+ // The entire `swift-numerics` API is now available
```

- Replace the use of "we" with "the Swift Numerics project". For example:

```diff
- Because we intend to make it possible to adopt Swift Numerics modules in the
- standard library at some future point...
+ The Swift Numerics project is intended to be adopted in the standard library
+ in future. Therefore, it...
```

- Provide three steps (`1... 2... 3...`) for the Using Swift Numerics in your project section, since it's a sequential process - this helps with UX
- Small refactor: add "that is"

```diff
- Swift Numerics is a standalone library separate from the core Swift project. In
+ Swift Numerics is a standalone library that is separate from the core Swift project. In
```

- Enhance several HowTo sections under "Contributing..." into discoverable subheadings (e.g. "to propose a new module" -> "# How to propose a new module")
- Add a "### Forums" section for improved discoverability
- Add "About" to enhance explanations of the links:

```
1. About [`RealModule`](Sources/RealModule/README.md)
    - [`ApproximateEquality`](Sources/RealModule/ApproximateEquality.swift)
2. About [`ComplexModule`](Sources/ComplexModule/README.md)
```

- Reformat the Markdown text to wrap at 80 characters (it's a good practice, just like in code)
- Turn complex sentences -> smaller easier to follow sentences
- Use second person ("you") instead of first person ("I", "we"), if you can
- Add blank lines between the section titles and first paragraphs
- Add blank lines between before and after code blocks

Let me know what you think.

Cheers!

markuswntr:
Thanks to  @8bitmp3, this PR now address not only the approximate equality module but also spelling/grammar - so I have changed the title to reflect the changes.

README.md
```diff
@@ -1,96 +1,154 @@
...
-However, there is also a top-level `Numerics` module that simply re-exports the complete public interface of swift-numerics:
+
+There is also a top-level `Numerics` module that re-exports the complete public
+interface of swift-numerics:
+
 ```swift
 import Numerics
 
-// All swift-numerics API is now available
+// The entire `swift-numerics` API is now available
```
stephentyrone, line 32: `swift-numerics` doesn't seem to be appropriate formatting here; "swift-numerics" is not code. In prose we usually style the project "Swift Numerics".

README.md
```diff
@@ -1,96 +1,154 @@
...
+Questions about how to use Swift Numerics modules, or issues that are not
+clearly bugs can be discussed in the ["Swift Numerics" section of the Swift
+forums](https://forums.swift.org/c/related-projects/swift-numerics).
 
 ## Modules
-1. [RealModule](Sources/RealModule/README.md)
-2. [ComplexModule](Sources/ComplexModule/README.md)
+
+1. About [`RealModule`](Sources/RealModule/README.md)
+    - [`ApproximateEquality`](Sources/RealModule/ApproximateEquality.swift)
```
stephentyrone, line 113: We should definitely link to detail on ApproximateEquality, but we should link to a markdown document, rather than the implementation. Let's remove this for now and open an issue to track adding one.

README.md
```diff
@@ -1,96 +1,154 @@
...
+
+Questions about how to use Swift Numerics modules, or issues that are not
+clearly bugs can be discussed in the ["Swift Numerics" section of the Swift
+forums](https://forums.swift.org/c/related-projects/swift-numerics).
 
 ## Modules
-1. [RealModule](Sources/RealModule/README.md)
-2. [ComplexModule](Sources/ComplexModule/README.md)
+
+1. About [`RealModule`](Sources/RealModule/README.md)
```
stephentyrone, line 112: I don't think "About" adds any clarity here (it actually confuses it). This is a list of modules, and the module is not "About RealModule" it is "RealModule".

8bitmp3: @stephentyrone Sounds good! Maybe we can have the title of the section named "# More on modules" or something along those lines? What each link is about/does can be quite important to increase clarity, better a11y, etc. WDYT?

README.md
8bitmp3: Agreed, thanks @stephentyrone

README.md
```diff
@@ -1,96 +1,154 @@
...
+type and module have the same name (discussion here:
+https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482). This would
+prevent users of Swift Numerics who don't need generic types from doing things,
+such as:
+
 ```swift
 import Complex
-// I know I only ever want Complex<Double>, so I shouldn't need the generic parameter.
-typealias Complex = Complex.Complex<Double> // doesn't work, because name lookup fails.
+// You only ever want Complex<Double>, so you shouldn't need the generic parameter.
```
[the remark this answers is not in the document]
8bitmp3: If we can use the second person ("you" - the reader/user) in docs instead of the first person ("I" and "we"), we should do so, according to Google and Microsoft style guides for doc writing. That's the reason for changing it here.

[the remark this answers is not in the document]
xwu: I think @stephentyrone’s point here is that the comment here is _not_ being written in the voice of the documentation writer, and what’s more essentially, _should not_ be expressed in those terms, since it is demonstrating particulars and not providing general advice.

README.md
xwu: I think the heading is fine as-is. It is a list of modules, not a section “on modules.”

xwu (commented):
Personally, I would prefer not adding hard line wraps to this document. It makes future editing much more finicky, and the diffs much less readable. Since it is not the prevailing style for this project to line-wrap Markdown files, I would suggest leaving that alone.

8bitmp3:
I understand that line wraps aren't too user-friendly @xwu They are part of the docs-as-code approach and help users see "the end of the line", since Markdown rendering can't always auto-wrap depending on where you view the files. This linting style has been adopted by certain well-known open source Google projects (they even include it in doc tests).

README.md
8bitmp3: I am just trying to follow commonly accepted guidelines of tech writing for good docs. 

The idea in our case is to help any user - whether they're new to Swift or Swift Numerics - to understand whether they should click on each link (this is part of the user experience/human-computer interface fields). So, in a way, this contributes to inclusiveness and accessibility, which are important to keep in mind when writing doc specs.

A few words with basic info can save readers a trip to an external site that may not want to go to (as per the Google style guide, as I couldn't find any relevant info on Microsoft or Apple's guides).

- For instance, this shows it's a README and I'd probably click on it, since READMEs are mostly common knowledge:

```
[`RealModule` - README](Sources/RealModule/README.md)
```

```
[About `RealModule`](Sources/RealModule/README.md)
```

- Compare this the original, where it's not too clear whether the user, who hasn't spent as much time with Swift Numerics as us, should click on the link. (Is it an API doc, a `.swift` file, a blog post, or an official Apple reference?)

```
[`RealModule`](Sources/RealModule/README.md)
```

Anyway, these are just suggestions. Maybe Apple's tech writers can help here.

Cheers!

README.md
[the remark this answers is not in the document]
8bitmp3: @xwu @stephentyrone Gotcha! Agreed

README.md
```diff
@@ -1,96 +1,154 @@
...
 
-Because we intend to make it possible to adopt Swift Numerics modules in the standard library at some future point, Swift Numerics uses the same license and contribution guidelines as the Swift project.
+Future expansion may assume the availability of other standard interfaces, such
+as [BLAS (Basic Linear Algebra
+Subprograms)](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms)
+and [LAPACK (Linear Algebra Package)](https://en.wikipedia.org/wiki/LAPACK). But
+modules with more specialized dependencies (or dependencies that are not
+available on all platforms supported by Swift) belong in a separate package.
+
+The Swift Numerics project is intended to be adopted in the standard library
```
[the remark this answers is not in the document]
stephentyrone: I'm actually the project owner, and @xwu is correct, the original text is correct, and this suggested change is not. There will surely be some components of Swift Numerics that never make their way into the standard library.

README.md
markuswntr: Thanks! I removed the link and opened an issue: https://github.com/apple/swift-numerics/issues/153

README.md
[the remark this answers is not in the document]
8bitmp3: Sounds good!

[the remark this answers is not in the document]
8bitmp3: I'll revert these changes back @stephentyrone - commit coming up @markuswntr.

Thanks for the feedback!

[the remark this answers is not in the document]
8bitmp3: Thanks @markuswntr for making the changes and @stephentyrone and @xwu for the feedback!

stephentyrone:
> They are part of the docs-as-code approach

I'm confused by this claim; as @xwu called out, wrapping at 80-cols makes diffs much larger than they should be for minor changes, which makes code review harder and more error-prone.

> Markdown rendering can't always auto-wrap

_Which_ renderers can't? Certainly whatever Github uses to render for the web does, and so does Xcode, which are by far the two renderers that will be most commonly used. Are there specific renderers that don't, which we expect to be frequently used with the Swift Numerics documentation (and are there reasons why viewing the docs on the web is not a viable option in those cases?)

8bitmp3:
@stephentyrone I understand where you and @xwu come from and the GitHub web UI helps to easily the entire text, while applying soft-wrapping and other user-friendly features.

I can revert the wrapping changes back so the diffs are more visible - whatever makes it easier for you to review here 👍

--- 

> Which renderers can't? 

I've reviewed code and docs in VSCode and used various IDEs - from IDEA to Xcode - and I'm aware some folks don't always have the default settings set to "auto-wrap": "on". However, many things in modern IDEs are automated by default these days, including the awesome Xcode.

So, for instance, if a contributor forks a repo, opens a Markdown file, and decides to make certain changes, they may see some long lines that make up long paragraphs, such as:

image

For some background, there's a GitLab blog post about the pros and cons of wrapping text, that summarizes this quite well:

image

https://about.gitlab.com/blog/2016/10/11/wrapping-text/

README.md
```diff
@@ -1,96 +1,117 @@
...
-Swift Numerics provides a set of modules that support numerical computing in Swift.
-These modules fall broadly into two categories:
+
+Swift Numerics provides a set of modules that support numerical computing in Swift. These modules fall broadly into two categories:
 
 - API that is too specialized to go into the standard library, but which is sufficiently general to be centralized in a single common package.
 - API that is under active development toward possible future inclusion in the standard library.
 
-There is some overlap between these two categories, and API that begins in the first category may migrate to the second as it matures and new uses are discovered.
+There is some overlap between these two categories, and an API that begins in the first category may migrate into the second as it matures and more users start using Swift Numerics.
```
8bitmp3, line 10: Originally: "...and new uses are discovered." was 100% good. Let's revert here

README.md
```diff
@@ -1,96 +1,117 @@
 # Swift Numerics
   
 ## Introduction
-Swift Numerics provides a set of modules that support numerical computing in Swift.
-These modules fall broadly into two categories:
+
+Swift Numerics provides a set of modules that support numerical computing in Swift. These modules fall broadly into two categories:
```
stephentyrone, line 5: While we don't want to wrap, we _do_ want one sentence per line.
```suggestion
Swift Numerics provides a set of modules that support numerical computing in Swift.
These modules fall broadly into two categories:
```

README.md
```diff
@@ -1,96 +1,117 @@
...
 
 - API that is too specialized to go into the standard library, but which is sufficiently general to be centralized in a single common package.
 - API that is under active development toward possible future inclusion in the standard library.
 
-There is some overlap between these two categories, and API that begins in the first category may migrate to the second as it matures and new uses are discovered.
+There is some overlap between these two categories, and an API that begins in the first category may migrate into the second as it matures and new uses are discovered.
+
+Swift Numerics modules are fine-grained. For example, if you need support for Complex numbers, you can import ComplexModule¹ as a standalone module:
```
stephentyrone, line 12:
```suggestion
Swift Numerics modules are fine-grained.
For example, if you need support for Complex numbers, you can import ComplexModule¹ as a standalone module:
```

README.md
```diff
@@ -1,96 +1,117 @@
...
-// All swift-numerics API is now available
+// The entire Swift Numerics API is now available
 ```
 
 Swift Numerics modules have minimal dependencies on other projects.
+
 The current modules assume only the availability of the Swift and C standard libraries and the runtime support provided by compiler-rt.
-Future expansion may assume the availability of other standard interfaces such as [BLAS (Basic Linear Algebra Subprograms)](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) and [LAPACK (Linear Algebra Package)](https://en.wikipedia.org/wiki/LAPACK), but modules with more specialized dependencies (or dependencies that are not available on all platforms supported by Swift) belong in a separate package.
+
+Future expansion may assume the availability of other standard interfaces, such as [BLAS (Basic Linear Algebra Subprograms)](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) and [LAPACK (Linear Algebra Package)](https://en.wikipedia.org/wiki/LAPACK). But modules with more specialized dependencies (or dependencies that are not available on all platforms supported by Swift) belong in a separate package.
```
stephentyrone, line 32:
```suggestion
Future expansion may assume the availability of other standard interfaces, such as [BLAS (Basic Linear Algebra Subprograms)](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) and [LAPACK (Linear Algebra Package)](https://en.wikipedia.org/wiki/LAPACK).
But modules with more specialized dependencies (or dependencies that are not available on all platforms supported by Swift) belong in a separate package.
```

README.md
```diff
@@ -1,96 +1,117 @@
...
-and finally, add `import Numerics` in your source code.
-    
+
+3. Add `import Numerics` in your source code.
+
 ## Contributing to Swift Numerics
-Swift Numerics is a standalone library separate from the core Swift project.
-In practice, it will act as a staging ground for some APIs that may eventually be incorporated into the Swift Standard Library, and when that happens such changes will be proposed to the Swift Standard Library using the established evolution process of the Swift project.
+
+Swift Numerics is a standalone library that is separate from the core Swift project. In practice, it will act as a staging ground for some APIs that may eventually be incorporated into the Swift Standard Library. When that happens, such changes will be proposed to the Swift Standard Library using the established evolution process of the Swift project.
```
stephentyrone, line 59:
```suggestion
Swift Numerics is a standalone library that is separate from the core Swift project, but it will sometimes act as a staging ground for APIs that will later be incorporated into the Swift Standard Library.
When that happens, such changes will be proposed to the Swift Standard Library using the established evolution process of the Swift project.
```

README.md
```diff
@@ -1,96 +1,117 @@
...
+1. [Large Fixed-Width Integers](https://github.com/apple/swift-numerics/issues/4)
+2. [Arbitrary-Precision Integers](https://github.com/apple/swift-numerics/issues/5)
+3. [Shaped Arrays](https://github.com/apple/swift-numerics/issues/6)
+4. [Decimal Floating-point](https://github.com/apple/swift-numerics/issues/7)
 
 ## Notes
-¹ Swift is currently unable to use the fully-qualified name for types when a type and module have the same name (discussion here: https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482).
-This would prevent users of Swift Numerics who don't need generic types from doing things like:
+
+¹ Swift is currently unable to use the fully-qualified name for types when a type and module have the same name (discussion here: https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482). This would prevent users of Swift Numerics who don't need generic types from doing things, such as:
```
stephentyrone, line 99:
```suggestion
¹ Swift is currently unable to use the fully-qualified name for types when a type and module have the same name (discussion here: https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482).
This would prevent users of Swift Numerics who don't need generic types from doing things, such as:
```

README.md
```diff
@@ -1,96 +1,117 @@
...
 // I know I only ever want Complex<Double>, so I shouldn't need the generic parameter.
 typealias Complex = ComplexModule.Complex<Double>
 // But I can still refer to the generic type by qualifying the name if I need it occasionally:
 let a = ComplexModule.Complex<Float>
 ```
-The `Real` module does not contain a `Real` type, but does contain a `Real` protocol, and users may want to define their own `Real` type (and possibly re-export the `Real` module), so the suffix is also applied there.
- New modules have to evaluate this decision carefully, but can err on the side of adding the suffix.
- It's expected that most users will simply `import Numerics`, so this isn't an issue for them.
+
+The `Real` module does not contain a `Real` type, but does contain a `Real` protocol, and users may want to define their own `Real` type (and possibly re-export the `Real` module) - that is why the suffix is also applied there. New modules have to evaluate this decision carefully, but can err on the side of adding the suffix. It's expected that most users will simply `import Numerics`, so this isn't an issue for them.
```
stephentyrone, line 117:
```suggestion
The `Real` module does not contain a `Real` type, but does contain a `Real` protocol.
Users may want to define their own `Real` type (and possibly re-export the `Real` module)--that is why the suffix is also applied there.
New modules have to evaluate this decision carefully, but can err on the side of adding the suffix.
It's expected that most users will simply `import Numerics`, so this isn't an issue for them.
```

README.md
8bitmp3: Thanks @stephentyrone - moving this sentence to the next paragraph LGTM

README.md
8bitmp3: Agreed. The proposed change looks good - will incorporate, thanks @stephentyrone

README.md
stephentyrone: To be clear, this is _not_ a new paragraph. New paragraphs only occur in markdown on a double line break.

README.md
8bitmp3: 👍 @stephentyrone Will make this change

README.md
8bitmp3: Agree with the rewording and splitting into more paragraphs 👍 Will make these changes

README.md
8bitmp3: 👍👍

README.md
8bitmp3: 👍👍 Thanks @stephentyrone

README.md
8bitmp3: @stephentyrone Got it!

README.md
```diff
@@ -1,96 +1,125 @@
...
+1. [Large Fixed-Width Integers](https://github.com/apple/swift-numerics/issues/4)
+2. [Arbitrary-Precision Integers](https://github.com/apple/swift-numerics/issues/5)
+3. [Shaped Arrays](https://github.com/apple/swift-numerics/issues/6)
+4. [Decimal Floating-point](https://github.com/apple/swift-numerics/issues/7)
 
 ## Notes
+
 ¹ Swift is currently unable to use the fully-qualified name for types when a type and module have the same name (discussion here: https://forums.swift.org/t/pitch-fully-qualified-name-syntax/28482).
-This would prevent users of Swift Numerics who don't need generic types from doing things like:
+This would prevent users of Swift Numerics who don't need generic types from doing things, such as:
```
xwu, line 104:
```suggestion
This would prevent users of Swift Numerics who don't need generic types from doing things such as:
```

README.md
```diff
@@ -1,96 +1,125 @@
...
+// The entire Swift Numerics API is now available
 ```
 
 Swift Numerics modules have minimal dependencies on other projects.
+
 The current modules assume only the availability of the Swift and C standard libraries and the runtime support provided by compiler-rt.
-Future expansion may assume the availability of other standard interfaces such as [BLAS (Basic Linear Algebra Subprograms)](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) and [LAPACK (Linear Algebra Package)](https://en.wikipedia.org/wiki/LAPACK), but modules with more specialized dependencies (or dependencies that are not available on all platforms supported by Swift) belong in a separate package.
+
+Future expansion may assume the availability of other standard interfaces, such as [BLAS (Basic Linear Algebra Subprograms)](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) and [LAPACK (Linear Algebra Package)](https://en.wikipedia.org/wiki/LAPACK).
+But modules with more specialized dependencies (or dependencies that are not available on all platforms supported by Swift) belong in a separate package.
```
xwu, line 35:
```suggestion
, but modules with more specialized dependencies (or dependencies that are not available on all platforms supported by Swift) belong in a separate package.
```

README.md
8bitmp3: Got it @xwu Thanks!

README.md
8bitmp3: Thanks for reviewing @xwu, appreciate it. If you think it's important to have this section as one long sentence, I can definitely make this change!

I noticed that in GitHub's web UI, the whole paragraph takes up three lines. If we merge these sentences, they become a one long thing spanning three lines:

image

---

Here's the reasoning behind simplifying a large sentence into two shorter ones (these are guidelines, not rules):

From Google Technical Writing - [Short sentences](https://developers.google.com/tech-writing/one/short-sentences):

> "Shorter documentation reads faster than longer documentation."
> 
> "Finding the shortest documentation implementation takes time but is ultimately worthwhile. Short sentences communicate more powerfully than long sentences, and short sentences are usually easier to understand than long sentences."

From Microsoft Style Guide - [Writing Tips](https://docs.microsoft.com/en-us/style-guide/global-communications/writing-tips):

> "These practices will help localizers and customers.
> "Write short, simple sentences. Punctuating a sentence with more than a few commas and end punctuation usually indicates a complex sentence. Consider rewriting it or breaking it into multiple sentences."

https://developers.google.com/tech-writing/one/short-sentences

xwu: This is one thought, not two, so it needs to be one sentence.

8bitmp3: Ok @xwu! Will make the changes

stephentyrone:
Hi @markuswntr, can you update this to point against main instead of master? Thanks.

markuswntr:
Hi @stephentyrone, updated base branch to main. Thanks!

stephentyrone (approved):

PR #155, opened 2020-09-16
Add complex log(onePlus:), update log()

stephentyrone:
Adapts @NevinBR's cute factorization of the real part here:

https://forums.swift.org/t/elementaryfunctions-compliance-for-complex/37903/3

Adds a little bit of FMA magic to avoid catastrophic cancellation and recover a good componentwise error bound as well as an error bound in the complex norm.


stephentyrone:
@NevinBR what would you prefer to be credited as in the comments?

stephentyrone:
Note that this doesn't yet handle Im(z) ≈ -1 right. That's relatively easy to fix up, but I have some other work to do today, so it'll have to wait.

NevinBR:
> @NevinBR what would you prefer to be credited as in the comments?

I’d be a little surprised if I’m the first to create an implementation like this, but any of “Nevin”, “NevinBR”, or “Nevin B-R” would be fine.

Sources/ComplexModule/ElementaryFunctions.swift
```diff
@@ -154,78 +154,136 @@ extension Complex /*: ElementaryFunctions */ {
...
+    // and -2), it is zero, so we need to be careful about cancellation.
+    //
+    // To begin, let's just see what domain z actually lies in.
+    let r2 = (1 + z).lengthSquared
+    // If |1+z| <= ½, we're well inside the circle (and close enough
+    // to -1 that 1 + z is computed exactly. If |1+z| => 2, we're far
+    // enough outside the circle that we can simply use log(1+z)--1+z
+    // may round, but that rounding doesn't cause problematic error in
+    // the final result.
+    if 2*r2 <= 1 && !(r2 < 2) { return log(1 + z) }
```
NevinBR, line 266: Shouldn’t this use boolean “or” instead of “and”?

And based on the paragraph before it, those 2’s should be 4’s because of squaring, right?

Sources/ComplexModule/ElementaryFunctions.swift
```diff
@@ -154,78 +154,136 @@ extension Complex /*: ElementaryFunctions */ {
...
-    // Unlike log(z), we do not need to worry about what happens if a
-    // underflows.
-    return Complex(
-      RealType.log(onePlus: a)/2,
-      RealType.atan2(y: z.y, x: 1+z.x)
-    )
+    //  	If x is in [-3, -1], everything is exact because of Sterbenz.
+    //    If x is in (-1, +1], 2 + x will generally round, but 2 > |x|,
+    //      so no sorting is needed to order the tail computation.
+    return log(1+z)
```
NevinBR, line 286: This doesn’t appear to use the careful calculation that was described in the comments.

stephentyrone: Yeah, I'm revising it to handle some other cases, but got side-tracked into improving log( ) first.

stephentyrone: (i.e. reviewing log(onePlus:) is wasted effort, I just hacked a merge together so it would build and I could do some testing on log( ).)

Issue #156, opened 2020-09-17
Unable to Compile with Xcode 12 (12A7209)


b3ll:
I'm running Catalina 10.15.6 (19G2021), yet because of `Float16` I'm unable to compile swift-numerics when targeting macOS.

Swift compiler is reporting the correct version `Apple Swift version 5.3 (swiftlang-1200.0.29.2 clang-1200.0.30.1)`, yet I don't understand why it's trying to compile `Float16` when it's explicitly marked as `unavailable` for macOS. Building for an iOS device or the iOS simulator works fine. I tried using `swift-numerics` with a sample Xcode project for a macOS app as well but it fails with the same issues.

Any idea what's up?

**swift build**
```bash
» xcrun swift build
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:18:11: error: cannot find type 'Float16' in scope
extension Float16: Real {
          ^~~~~~~
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:20:43: error: cannot find type 'Float16' in scope
  public static func cos(_ x: Float16) -> Float16 {
                                          ^~~~~~~
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:20:31: error: cannot find type 'Float16' in scope
  public static func cos(_ x: Float16) -> Float16 {
                              ^~~~~~~
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:25:43: error: cannot find type 'Float16' in scope
  public static func sin(_ x: Float16) -> Float16 {
                                          ^~~~~~~
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:25:31: error: cannot find type 'Float16' in scope
  public static func sin(_ x: Float16) -> Float16 {
                              ^~~~~~~
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:30:43: error: cannot find type 'Float16' in scope
  public static func tan(_ x: Float16) -> Float16 {
                                          ^~~~~~~
/Users/Adam/Downloads/swift-numerics/Sources/RealModule/Float16+Real.swift:30:31: error: cannot find type 'Float16' in scope
  public static func tan(_ x: Float16) -> Float16 {
                              ^~~~~~~
[...]

PR #157, opened 2020-09-18
Swap availability for #if check on Float16.

stephentyrone:
Xcode 12 on Catalina has Swift 5.3 but the macOS 10.15 SDK, which [breaks the conditionals around Float16](https://github.com/apple/swift-numerics/issues/156). Since the type is unconditionally unavailable on macOS, we can simply `#if`-out any Float16 extensions instead.

### DIFF
--- a/Sources/RealModule/Float16+Real.swift
+++ b/Sources/RealModule/Float16+Real.swift
@@ -9,12 +9,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.3)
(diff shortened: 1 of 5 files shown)


lorentey (approved):
Looks good to me! (I'd probably leave the macOS availability annotations in.)

We really need some sort of compile-time way to check the version of the stdlib. We could have Package.swift do an `if #available macOS(11.0)` check and set a HAVE_FLOAT16 compile-time conditional based on that, but that seems a bit too fancy.

stephentyrone:
@lorentey yeah, step 2 here is definitely to figure out how to avoid this mess in the future =)

PR #159, opened 2020-10-12
build: enable building `Numerics` on Darwin with CMake

compnerd:
The `-enable-force-autolink` flag is not supported on Darwin.
Fortunately, it is also unnecessary to work around the SR on Darwin.
Simply exclude the flag when targeting Darwin.


stephentyrone (approved):
Thanks, this seems fine.

PR #172, opened 2021-02-19
Adjust Float16 availability for Swift 5.4 toolchains.

stephentyrone:
In older toolchains, Float16 is marked unavailable on all mac targets; starting with the 5.4 toolchain, it is available starting in macOS 11.0 when targeting Apple Silicon.


stephentyrone:
@compnerd can you sanity-check this on windows when you have a chance?

PR #178, opened 2021-03-19
Remove .xcodesamplecode.plist, which prevents editing markdown from Xcode

kylemacomber:
This is kind of cool, but it prevents editing the markdown from Xcode (but not any other text editor).


stephentyrone:
@swift-ci test

PR #181, opened 2021-08-19
Document the Augmented enum in RealModule in advance of 1.0

stephentyrone:


stephentyrone:
@swift-ci test

stephentyrone:
Cherry-picked to release/1.0.

Sources/RealModule/AugmentedArithmetic.swift
```diff
@@ -9,18 +9,85 @@
...
+  ///
+  /// `head` is the correctly rounded value of `a*b`. If no overflow or
+  /// underflow occurs, `tail` represents the rounding error incurred in
+  /// computing `head`, such that the exact product is the sum of `head`
+  /// and `tail` computed without rounding.
+  ///
+  /// This operation is sometimes called "twoProd" or "twoProduct".
+  ///
+  /// Edge Cases:
+  /// -
```
varungandhi-apple, line 31: There is a blank bullet here and also in some lists below. Is it intentional?

stephentyrone: Yes, this produces the desired formatting in Xcode: 
Screen Shot 2021-08-19 at 2 21 14 PM

Sources/RealModule/AugmentedArithmetic.swift
```diff
@@ -9,18 +9,85 @@
...
+  /// - When `head` is close to the underflow boundary, the rounding error
+  ///   may not be representable due to underflow, and `tail` will be rounded.
+  ///   If `head` is very small, `tail` may even be zero, even though the
+  ///   product is not exact.
+  /// - If `head` is zero, `tail` is also a zero with unspecified sign.
+  ///
+  /// Postconditions:
+  /// -
+  /// - If `head` is normal, then `abs(tail) < abs(head.ulp)`.
+  ///   Assuming IEEE 754 default rounding, `abs(tail) <= abs(head.ulp)/2`.
```
NevinBR, line 44: Are the “abs” on the right-hand sides necessary? I was under the impression that `ulp` was guaranteed to be non-negative for normal values. If so, the same thing applies to the doc comment for `sum(large:small:)`.

stephentyrone: That's right.

PR #182, opened 2021-08-19
Add some documentation for AlgebraicField ops, minor fix for Augmented.

stephentyrone:
Adds more specialized documentation for `reciprocal` on `Real`, because some of what's discussed for `AlgebraicField` does not apply, and we can simplify the discussion somewhat.


stephentyrone:
@NevinBR fix for the nit you flagged here.

stephentyrone:
@swift-ci test

PR #185, opened 2021-08-20
Further expand documentation of `reciprocal` with some example use.

stephentyrone:
Also added discussion of errors introduced by using reciprocal for real and complex types.

### DIFF
--- a/Sources/ComplexModule/Arithmetic.swift
+++ b/Sources/ComplexModule/Arithmetic.swift
@@ -52,13 +52,31 @@ extension Complex: AdditiveArithmetic {
 // and turn these into operators if/when we have it.
 // (https://github.com/apple/swift-numerics/issues/12)
 extension Complex {
+  /// `self` scaled by `a`.
   @usableFromInline @_transparent
   internal func multiplied(by a: RealType) -> Complex {
+    // This can be viewed in two different ways, which are mathematically
+    // equivalent: either we are computing `self * Complex(a)` (i.e.
+    // converting `a` to be a complex value, and then using the complex
+    // multiplication) or we are using the scalar product of the vector
+    // space structure: `Complex(a*real, a*imaginary)`.
+    //
+    // Although these two interpretations are _mathematically_ equivalent,
+    // they will generate different representations of the point at
+    // infinity in general. For example, suppose `self` is represented by
+    // `(infinity, 0)`. Then `self * Complex(1)` would evaluate as
+    // `(1*infinity - 0*0, 0*infinity + 1*0) = (infinity, nan)`, but
+    // the vector space interpretation produces `(infinity, 0)`. This does
+    // not matter much, because these are two representations of the same
+    // semantic value, but note that one requires four multiplies and two
+    // additions, while the one we use requires only two real multiplications.
     Complex(x*a, y*a)
   }
   
+  /// `self` unscaled by `a`.
   @usableFromInline @_transparent
   internal func divided(by a: RealType) -> Complex {
+    // See implementation notes for `multiplied` above.
     Complex(x/a, y/a)
   }
 }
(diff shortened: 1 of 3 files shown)


stephentyrone:
@swift-ci test

Sources/ComplexModule/Arithmetic.swift
```diff
@@ -147,17 +165,23 @@ extension Complex: AlgebraicField {
...
   ///   // Fallback on using division.
   ///   return data.map { $0 / divisor }
   /// }
   /// ```
+  ///
+  /// Error Bounds:
+  /// -
+  /// Unlike real types, when working with complex types multiplying by the
+  /// reciprocal instead of dividing cannot change the result. If the
+  /// reciprocal is non-nil, the two computations always always equivalent.
```
stephentyrone, line 184:
```suggestion
  /// reciprocal is non-nil, the two computations are always equivalent.
```

Sources/ComplexModule/Arithmetic.swift
```diff
@@ -147,17 +165,23 @@ extension Complex: AlgebraicField {
...
   ///     return data.map { $0 * recip }
   ///   }
   ///   // Fallback on using division.
   ///   return data.map { $0 / divisor }
   /// }
   /// ```
+  ///
+  /// Error Bounds:
+  /// -
+  /// Unlike real types, when working with complex types multiplying by the
```
xwu, line 182:
```suggestion
  /// Unlike real types, when working with complex types, multiplying by the
```

Sources/RealModule/AlgebraicField.swift
```diff
@@ -69,6 +69,30 @@ public protocol AlgebraicField: SignedNumeric {
   /// Note that `.zero.reciprocal`, somewhat surprisingly, is *not* nil
   /// for `Real` or `Complex` types, because these types have an
   /// `.infinity` value that acts as the reciprocal of `.zero`.
+  ///
+  /// If b.reciprocal is non-nil, you may be able to replace division by b
```
xwu, line 73:
```suggestion
  /// If `b.reciprocal is non-nil`, you may be able to replace division by `b`
```

Sources/RealModule/Real.swift
```diff
@@ -104,6 +104,46 @@ extension Real {
   ///
   /// If `a` if finite and nonzero, and `1/a` overflows or underflows,
   /// then `a.reciprocal` is `nil`. Otherwise, `a.reciprcoal` is `1/a`.
```
xwu, line 106:
```suggestion
  /// then `a.reciprocal` is `nil`. Otherwise, `a.reciprocal` is `1/a`.
```

Sources/RealModule/Real.swift
```diff
@@ -104,6 +104,46 @@ extension Real {
...
+  ///
+  /// If b.reciprocal is non-nil, you may be able to replace division by b
```
xwu, line 108:
```suggestion
  /// If `b.reciprocal` is non-nil, you may be able to replace division by `b`
```

Sources/RealModule/Real.swift
```diff
@@ -104,6 +104,46 @@ extension Real {
...
+  /// -
+  /// Multiplying by the reciprocal instead of dividing will slightly
+  /// perturb results. For example `5.0 / 3` is 1.6666666666666667, but
+  /// `5.0 * 3.reciprocal!` is 1.6666666666666665.
+  ///
+  /// The error of a normal division is bounded by half an ulp of the
+  /// result; we can derive a quick error bound for multiplication by
+  /// the real reciprocal (when it exists) as follows (I will use circle
+  /// operators to denote real-number arithmetic, and normal operators
+  /// for floating-point arithmetic:
```
xwu, line 136:
```suggestion
  /// for floating-point arithmetic):
```

stephentyrone:
Thanks @xwu!

PR #187, opened 2021-08-25
Remove Real and Complex _libraries_; keep modules.

stephentyrone:
These are vestigial, and not needed to support swift-pm.


stephentyrone:
@swift-ci please test

Issue #10, opened 2019-07-01
Integer utilities module


stephentyrone:
A new module providing additional functionality on the standard library integer types (and more generally, any fixed-width integer type), including:

- [x] - bitwise rotation
- [ ] - full add / sub (with carryin/carryout)
- [x] - saturating arithmetic (functions, not operators)
- [x] - division with rounding control
- [x] - shift with rounding control
- [x] - "true" mod function (Euclidean division)
- [x] - GCD
- [ ] - division by constant (precompute magic multiply and shift a la libdivide)
- [ ] - modulus by constant (use Lemire's method)

SusanDoggie:
Is bit reverse fit on this?

The FFT algorithm required bit reversal
https://en.wikipedia.org/wiki/Bit-reversal_permutation
https://github.com/apple/swift-numerics/blob/796a0dc3f74a60f102937f61cc363644513671d1/Sources/Accelerate/bit_reverse.swift

stephentyrone:
Definitely!

NevinBR:
Would basic number-theory functions be in scope for this module?

* gcd
* lcm
* Bézout coefficients
* Modular inverse

stephentyrone:
Yes, absolutely.

Lancelotbronner:
Is it possible to add it as a product and make a new release for it? I'd like to use the saturating arithmetic in embedded projects.

PR #189, opened 2021-08-26
GCD function for BinaryInteger

NevinBR:
<s>Added a `gcd` static method to `BinaryInteger`.</s>

(Edit: `gcd` is now a free function.)

The initial implementation uses the [binary GCD algorithm](https://en.wikipedia.org/wiki/Binary_GCD_algorithm).

I believe that `gcd` is a sufficiently widespread term of art to prefer it over a more verbose spelling. However if others disagree, we could use `greatestCommonDivisor` instead.


Sources/IntegerUtilities/GCD.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+extension BinaryInteger {
+  
+  /// The greatest common divisor of `a` and `b`.
+  ///
+  /// If both inputs are zero, the result is zero.
+  @inlinable
+  public static func gcd(_ a: Self, _ b: Self) -> Self {
```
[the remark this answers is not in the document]
xwu: I go back and forth on this a bit but mildly favor the static method as it is over the free function.

As to spelling, `gcd` is just *much* more readable than `greatestCommonDivisor`; it takes significantly longer to read and understand what's going on with the latter spelling.

Sources/IntegerUtilities/GCD.swift
```diff
@@ -0,0 +1,47 @@
...
+    // The binary GCD algorithm
+    //
+    // At the top of the loop both x and y are odd. Each pass removes at least
+    // one low-order bit from the larger of the two, so the number of iterations
+    // is bounded by the sum of the bit-widths of the inputs.
+    while x != 0 {
+      if x < y { swap(&x, &y) }
+      x -= y
+      x >>= x.trailingZeroBitCount
+    }
```
NevinBR, line 43: Thinking about micro-optimizations, should the loop be rearranged like this?

```
repeat {
  x >>= x.trailingZeroBitCount
  if x < y { swap(&x, &y) }
  x -= y
} while x != 0
```

It saves one `x != 0` (on the first pass), and allows removing the earlier line `x >>= xtz` to save one shift operation (on the last pass when `x` is zero).

The only drawback I see is that the explanatory comment would need to specify a location other than the top of the loop for where both values are odd, but that’s not really a problem.

Sources/IntegerUtilities/GCD.swift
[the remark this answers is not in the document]
NevinBR: Playing around with code that uses `gcd`, I agree that’s it’s a much better spelling than having the words written out in full.

I also find that needing to write out the name of the type makes the code less readable, in both simple and complex expressions.

Notably, the following should “just work”, and the programming experience is degraded if it doesn’t:

```swift
let g = gcd(m, n)
```

So unless anyone has a strong reason not to, I’m going to move `gcd` out to a free function.

Sources/IntegerUtilities/GCD.swift
stephentyrone: I am not really convinced that this is even an optimization; the compiler can basically do this for you, because x == 0 has already been checked above (so the compiler knows statically that x != 0 and will skip the first check), and people are generally more familiar with while than repeat-while, so there's some advantage. You can do _half_ of the change and eliminate the shift, I guess:

```
while x != 0 {
  x >>= x.trailingZeroBitCount
  if x < y { swap(&x, &y) }
  x -= y
}
```

stephentyrone:
One higher-level question; should we expose a gcd operation on > 2 integers, or on collections? I'm not sure how valuable a variadic gcd would be (since there's not a great way to marshal multiple arguments like that), but is it worth having a more convenient spelling than `reduce(into: 0) { $0 = gcd($0, $1) }`?

Sources/IntegerUtilities/GCD.swift
NevinBR: True, though I must say it’s not at all obvious “how smart” the compiler is, or what optimizations are guaranteed to occur. I think for now I’m going to leave the simple, symmetric version of the code as-is. If benchmarks reveal ways to make it more efficient, that can always be done later.

Another potential optimization would be, if either `x` or `y` equals 1 after the initial right-shifts, then we can just return `1 << min(xtz, ytz)`. Of course that check might slow things down the rest of the time, so it’s another place for benchmarking. I think we can safely leave the micro-optimizations for the future.

stephentyrone: The optimization in question here is absolutely trivial and does not require any real compiler smarts.

NevinBR: All right, I made `gcd` a free function, and then (in a separate commit for ease of diff-viewing) moved the right-shift to the top of the loop and eliminated the two earlier lines that are consequently unnecessary.

NevinBR:
> One higher-level question; should we expose a gcd operation on > 2 integers, or on collections?

That’s a good question. We should probably hash out things like that over on issue #10.

NevinBR:
Apropos of nothing, and not really relevant or important, but where the descriptive comment for the while loop states:

> Each pass removes at least one low-order bit from the larger of the two, so the number of iterations is bounded by the sum of the bit-widths of the inputs.

If I’ve done my analysis correctly then, letting m and n be, respectively, the smaller and larger bit-widths of x and y (after eliminating the trailing zeros), a tighter bound for the number of iterations is given by:

`n + (log(2)/log(𝜑) - 1)m`

Where 𝜑 = (1+√5)/2 is the golden ratio. So the bound is approximately `n + 0.44m`, which occurs when the ratio between x and y is near 𝜑 as their roles swap. This is found by accounting for the high-order bits which are eliminated in the subtraction when x and y have the same width.

I do not intend to change the text of the comment.

stephentyrone:
> We should probably hash out things like that over on issue #10.

I'm not sure #10 is the right place to discuss it, but I'm perfectly happy to punt it to another PR; it would clearly be built on top of this anyway.

Sources/IntegerUtilities/GCD.swift
```diff
@@ -0,0 +1,42 @@
...
+/// If both inputs are zero, the result is zero. If one input is zero, the
+/// result is the absolute value of the other input.
+///
+/// [wiki]: https://en.wikipedia.org/wiki/Greatest_common_divisor
+@inlinable
+public func gcd<T: BinaryInteger>(_ a: T, _ b: T) -> T {
+  var x = a.magnitude
+  var y = b.magnitude
+  
+  if y == 0 { return T(x) }
```
stephentyrone, line 23: We need to define what happens for `gcd(0, Int.min)`. The implementation here will trap attempting to construct `Int(.min.magnitude)`, which seems undesirable.

Three options:
1. Make the return type `T.Magnitude` instead of `T`. Always defined, but inconvenient in expressions.
2. Define `gcd(0, .min)`, `(.min, 0)` and `(.min, .min)` to return `.min`.
3. Trap.

I think I'm weakly in favor of (2), but open to convincing otherwise.

NevinBR:
> I'm not sure #10 is the right place to discuss it, but I'm perfectly happy to punt it to another PR; it would clearly be built on top of this anyway.

I was thinking along the lines of discussing the scope of the integer utilities modules and general shape of its APIs.

One thing I’ve found with Github is that it doesn’t really provide an obvious location for discussion, planning, and questions *about* a project.

Sources/IntegerUtilities/GCD.swift
NevinBR: Swift has consistently taken the view that integer types model the ring of integers, ℤ, and mathematical operations behave accordingly. If the result of an operation does not fit in the destination type, the program traps.

That’s why I wrote the code it this way: the GCD is well-defined and non-negative for all inputs, and if the result overflows there is a trap. This is no different from the existing subtraction operator, eg. `0 - .min`.

That would suggest `gcd` should stay as-is, and if a programmer wants to guarantee it won’t trap then they can either test for `.min` first, or work with `Magnitude` via:

```
let g = gcd(a.magnitude, b.magnitude)
```

We could also add documentation similar to that for [`SignedNumeric.negate()`](https://github.com/apple/swift/blob/a375ffebcf40e9f3392e297c24afd73c2aeb3623/stdlib/public/core/Integers.swift#L303-L319) regarding `.min` to warn of the trap, however it’s worth noting that the [`BinaryInteger` subtraction operator](https://github.com/apple/swift/blob/a375ffebcf40e9f3392e297c24afd73c2aeb3623/stdlib/public/core/Integers.swift#L832-L852) has nothing of the sort.

• • •

That said, I’m not totally opposed to (2), and I do recognize the value in being able to calculate first and then check for an edge-case result after, rather than preemptively needing to check for every possible input that could trap.

Still, it seems highly unusual that a greatest common divisor calculation would produce a negative result.

stephentyrone: That's fine; it should be documented and test cases added.

stephentyrone (dismissed):

stephentyrone:
@swift-ci test

stephentyrone:
Thanks @NevinBR! Do you want to do anything further with this before I merge it?

Sources/IntegerUtilities/GCD.swift
NevinBR: Okay I added documentation and tests for the overflow behavior.

…wait, I think `XCTExpectFailure` has higher version requirements than this package. Is there a different way the tests should be written?

stephentyrone: I was considering bumping the supported version for other reasons anyway post-1.0, so let me investigate that a little bit further first.

NevinBR: Is there a way to wrap the `expectFailure` tests in a conditional `#if` block?

stephentyrone: Let's remove them for now. Annoying, but we'll deal with it.

Sources/IntegerUtilities/GCD.swift
```diff
@@ -0,0 +1,49 @@
...
+/// The greatest common divisor of `a` and `b`.
+///
+/// If both inputs are zero, the result is zero. If one input is zero, the
+/// result is the absolute value of the other input.
+///
+/// The resulting value must be representable within the value's type. In
+/// particular, the gcd of a signed, fixed-width integer type's minimum
+/// with itself or zero results in a value that cannot be represented:
+///
+///     gcd(Int.min, Int.min)   // Overflow error
```
stephentyrone, line 21: Can we amend this to: "the gcd of a signed, fixed-width integer type's minimum with itself or zero cannot be represented, and results in a trap."

NevinBR: I’m fine with that, though I have a slight preference for retaining the colon at the end rather than turning it into a period.

The existing wording was chosen simply to be consistent with [SignedNumeric.negate()](https://github.com/apple/swift/blob/a375ffebcf40e9f3392e297c24afd73c2aeb3623/stdlib/public/core/Integers.swift#L312-L319), except for the colon which I added.

Though reading it again, in the sentence prior to what you quoted, I think the use of “the value” is unclear. For `negate`, “the value” presumably refers to the instance on which the method was called, but `gcd` is not an instance method.

Perhaps it should become either “The resulting value must be representable.” or “The resulting value must be representable within its type.”

NevinBR: Okay, I made a change, let me know what you think.

I went with your wording, including the final period instead of colon, except I added parentheses around “or zero”. I can remove them if you prefer.

I also simplified the previous sentence.

Sources/IntegerUtilities/GCD.swift
NevinBR: Okay, I commented out the `ExpectFailure` tests, with a `TODO` to enable them when possible.

If you’d prefer I remove them entirely, just let me know.

stephentyrone (approved):

stephentyrone:
@swift-ci test

stephentyrone:
Thanks for your patience, @NevinBR!

PR #198, opened 2021-08-31
Clarify that differentaible API is not stable.

stephentyrone:
@rxwei reminded me that differentiable isn't source stable, so API involving it cannot be, by definition. Clarify this in the README.


stephentyrone:
@swift-ci test

PR #203, opened 2021-09-26
Skeleton of integer shift with rounding.

stephentyrone:
Fairly correct implementation of right-shift with rounding. Also fairly efficient, though I'm sure I'll find some cases where we're generating uncessary overflow checks upon further inspection. Nonetheless, a decent enough start.

### DIFF
--- a/Sources/IntegerUtilities/Rounding.swift
+++ b/Sources/IntegerUtilities/Rounding.swift
@@ -1,0 +1,139 @@
+//===--- Rounding.swift ---------------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+public enum RoundingRule {
+  case down
+  case up
+  case towardZero
+  case awayFromZero
+//     toEven
+  case toOdd
+//     toNearestOrDown
+//     toNearestOrUp
+//     toNearestOrTowardZero
+  case toNearestOrAwayFromZero
+  case toNearestOrEven
+//     toNearestOrOdd
+  case stochastic
+  case trap
+}
+
+extension BinaryInteger {
+  public func shifted<Count: BinaryInteger>(
+    right count: Count,
+    rounding rule: RoundingRule = .down
+  ) -> Self {
+    // Easiest case: count is zero or negative, so shift is always exact;
+    // delegate to the normal >> operator.
+    if count <= 0 { return self >> count }
+    if count >= bitWidth {
+      // Note: what follows would cause an infinite loop if bitWidth <= 1.
+      // This will essentially never happen, but in the highly unlikely event
+      // that we encounter such a case, we promote to Int8, do the shift, and
+      // then convert back to the appropriate result type.
+      if bitWidth <= 1 {
+        return Self(Int8(self).shifted(right: count, rounding: rule))
+      }
+      // That pathological case taken care of, we can now handle over-wide
+      // shifts by first shifting all but bitWidth - 1 bits with sticky
+      // rounding, and then shifting the remaining bitWidth - 1 bits with
+      // the desired rounding mode.
+      let count = count - Count(bitWidth - 1)
+      var floor = self >> count
+      let frac = self - (floor << count)
+      if frac != 0 { floor |= 1 } // insert sticky bit
+      return floor.shifted(right: bitWidth - 1, rounding: rule)
+    }
+    // Now we are in the happy case: 0 < count < bitWidth, which makes all
+    // the math to handle rounding simpler.
+    //
+    // TODO: If we were really, really careful about overflow, some of these
+    // could be made simpler. E.g. mathematically round up is implemented here
+    // via:
+    //
+    //   floor = self >> count
+    //   mask = 1 << count - 1
+    //   frac = self & mask
+    //   return floor + (frac == 0 ? 0 : 1)
+    //
+    // _if_ we didn't have to worry about intermediate overflow (either because
+    // self is bounded away from .max or because we don't have a fixed-width
+    // type), then we could use the following instead:
+    //
+    //   mask = 1 << count - 1
+    //   return (self + mask) >> count
+    //
+    // However, self + mask can overflow, e.g. if we're shifting .max by 1:
+    //
+    //   mask = 1 << 1 - 1 = 1
+    //   self + mask = .max + 1 // uh-oh
+    //
+    // This cannot occur for arbitrary-precision numbers, so we could use
+    // the simpler expressions for those (but those are precisely the cases
+    // where performance does not matter). More interesting, we could promote
+    // fixed-width numbers to a wider type, preventing this intermediate
+    // overflow and allowing us to use the simpler expressions much of the
+    // time. We could also explicitly _detect_ the overflow if it happens and
+    // patch up the result, though this is a little bit tricky because
+    // addingReportingOverflow does not exist on BinaryInteger. Hence, this
+    // is a TODO for the future.
+    let mask = Magnitude(1) << count - 1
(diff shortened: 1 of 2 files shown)


stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

Sources/IntegerUtilities/Rounding.swift
```diff
@@ -0,0 +1,65 @@
...
+// Copyright (c) 2021 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+/// A rule that defines how to select one of the two representable results
+/// closest to a given value.
+public enum RoundingRule {
```
Catfish-Man, line 14: This feels a little inconsistent.

`…, rounding: .down` ✅ 
`..., rounding: .stochastic` should this be "stochastically"?
`..., rounding: .trap` in what circumstances does it trap? (I realize this is clear if you think about it for a moment, but maybe we can make it obvious at the point of use?)

Related, do we want `rounding` or do we want `rounded`? With the name `shifted` we're describing the resulting value, rather than the action, but with `rounding` we're describing the action.

Sources/IntegerUtilities/Shift.swift
```diff
@@ -0,0 +1,149 @@
...
+  ///     7.shifted(right: 2, rounding: .stochastic)
+  ///
+  ///     // 4/2^2 = 4/4 = 1, exactly.
+  ///     4.shifted(right: 2, rounding: .trap)
+  ///
+  ///     // 5/2 is 2.5, which is not exact, so this traps.
+  ///     5.shifted(right: 1, rounding: .trap)
+  @inlinable
+  public func shifted<Count: BinaryInteger>(
+    right count: Count,
```
Catfish-Man, line 41: I forget, what's the precedent here: do we usually include or omit "By"?

Sources/IntegerUtilities/Shift.swift
```diff
@@ -0,0 +1,149 @@
...
+    rounding rule: RoundingRule = .down
```
Catfish-Man, line 42: As someone who doesn't think about rounding very often I'm just going to assume that you've thought about this and there's an obvious default here, but I'm leaving this comment just on the tiny chance that on seeing it again you'll go "hm actually maybe they should always have to say what they want" :)

Catfish-Man (commented):
Reviewed API only, didn't look at the implementation at all

Sources/IntegerUtilities/Rounding.swift
```diff
@@ -0,0 +1,65 @@
...
+  /// Produces the closest representable value that is less than or equal
+  /// to the value being rounded.
+  ///
+  /// This is the default rounding mode for integer shifts, including the
+  /// shift operators defined in the standard library.
+  case down
```
milseman, line 20: What do you think about additionally providing something like:

```swift
static var `default`: Self { .down }
```

Sources/IntegerUtilities/Rounding.swift
stephentyrone: Yeah, I'm not super-delighted with "trap" either. Still trying to come up with a better name for it.

Sources/IntegerUtilities/Rounding.swift
stephentyrone: I think the main concern would be that we might add an operation someday that wants to use a different default. round-to-nearest multiple comes to mind as a possibility.

milseman: Ah, so default depends on operation (so either current design or name it `defaultShifting`, etc.). What about a "fastest" or "I don't care" computed var, would that also be dependent on the operation?

Sources/IntegerUtilities/Shift.swift
```diff
@@ -0,0 +1,149 @@
...
+      return floor + Self((round + frac) >> count)
+    case .toNearestOrEven:
+      let round = mask >> 1 + Magnitude(floor & 1)
+      return floor + Self((round + frac) >> count)
+    case .stochastic:
+      // TODO: it's unfortunate that we can't specify a custom random source
+      // for the stochastic rounding rule, but I don't see a nice way to have
+      // that share the API with the other rounding rules, because we'd then
+      // have to take the RNG in-out. The same problem applies to rounding
+      // with dithering. We should consider adding a stateful rounding API
```
glessard, line 124: Is mentioning dithering relevant? I find it confusing, especially since I thought that it is exactly this case.

stephentyrone: The `//` comments are mostly notes for myself, rather than public documentation. =)

Sources/IntegerUtilities/Rounding.swift
rxwei: `trapIfUnrepresentable`?

Sources/IntegerUtilities/Shift.swift
tbkka: For right shift, the default should certainly be to truncate.  That's `down` or `towardsZero` for positive values.  I have to stop and think for a bit before I can remember the right behavior for negative values.

Sources/IntegerUtilities/Shift.swift
tbkka: I'm not sure I'd include stochastic rounding in this API.  It raises too many questions.  (Ulichney's "Digital Halftoning" is a great introduction that discusses the importance of controlling the frequency distribution of your noise source.)

stephentyrone: Blue noise and ordered dithers require a stateful rounding API, but even stochastic rounding without state is quite useful.

Sources/IntegerUtilities/Rounding.swift
xwu: `.exactOrTrap` maybe?

Sources/IntegerUtilities/Shift.swift
stephentyrone: Right; "normal" right shifts (>>) round down, and there's no good reason to make anything else the default. "Normal" division (/) rounds toward zero, which is why divide by power of two can't be replaced with a shift for signed values, but the way to resolve that mismatch is to make the forthcoming divide-with-rounding default to .down as well, as that gives a more sensible remainder for most uses.

Sources/IntegerUtilities/Rounding.swift
stephentyrone: Maybe just `.exact`, though I would like to mention "trap" in the name. I like the symmetry of `exact` with the `exactly:` inits from the stdlib.

tbkka: I like .exactOrTrap  or something similar.  "Trap" is kind of a specialist word, though, so maybe .exactOrFatal?  .exactOrAssert?  .mustBeExact?  I think it makes sense to have this one be linguistically distinct, so .onlyExact or .mustBeExact or .remainderMustBeZero or ...

benrimmington: If this was implemented in the standard library, would the new cases be added to [`FloatingPointRoundingRule`](https://developer.apple.com/documentation/swift/floatingpointroundingrule), or should the new enum be renamed to `IntegerRoundingRule` instead?

stephentyrone:
@swift-ci test

Sources/IntegerUtilities/Rounding.swift
stephentyrone: "don't care" here would just be `>>`, and for divide it would be "/". Expectation is that people reach for these when they _do_ care.

Sources/IntegerUtilities/Rounding.swift
stephentyrone: I would like to deprecate `FloatingPointRoundingRule` eventually and replace it with `RoundingRule`. Exactly what that migration path looks like isn't totally clear, but I think it's the direction that we'll take if/when these show up in the stdlib.

PR #205, opened 2021-10-07
Account for overflow/underflow in Quaternion `length`

markuswntr:
Given that `lengthSquared` of Quaternion is fairly prone to overflow/underflow, `length` (= `sqrt(lengthSquared)`) could yield an invalid result even if it would be perfectly representable.
This PR fixes the problem in that it handles overflow or underflow of `lengthSquared` when calculating `length`.


stephentyrone:
Markus, please update to resolve the conflicts and I'll merge.

PR #208, opened 2021-10-21
Add `Augmented.sum(_:_:)` aka "twoSum"

markuswntr:
With augmented arithmetic in place, I was wondering if we might as well add "twoSum" alongside the existing "fast2sum".
https://en.wikipedia.org/wiki/2Sum

As with "fast2sum", which is currently used for the elementary function of complex numbers, "twoSum" would be useful for the elementary function of quaternions where, due the additional pair of imaginary numbers, the magnitudes of the summands can not be guaranteed to be in any order.


stephentyrone:
Mind adding a couple test cases?

PR #209, opened 2021-11-03
Remove big/little endiant distinction in representation of DoubleWidth

stephentyrone:
This doesn't really add anything that we need; it changes the in-memory layout, but that's effectively an implementation detail. We might as well simply pick one and simplify things.

This does mean that DoubleWidth types have a "weird" layout on big endian systems, but that doesn't really cause any problems for anything. This _would_ be a problem if we sunk these types into the standard library and wanted to bridge them to/from `__int128_t` and friends, because the representation wouldn't match. It's also trivial to revert this change, however, since it doesn't actually intrude into any operations, so let's keep things simple for now.

### DIFF
--- a/Sources/_TestSupport/DoubleWidth.swift
+++ b/Sources/_TestSupport/DoubleWidth.swift
@@ -47,11 +47,7 @@ public struct DoubleWidth<Base : FixedWidthInteger> {
   public typealias High = Base
   public typealias Low = Base.Magnitude
 
-#if _endian(big)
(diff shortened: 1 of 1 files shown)


stephentyrone:
@swift-ci please test

stephentyrone:
@swift-ci please test

xwu:
Fiiine. I was really pleased with my cleverness here.

A point in favor of simplification here is that even on little endian systems there can still be an alignment mismatch between this and the non-`DoubleWidth` types of the same width.

PR #210, opened 2021-11-03
Expand some comments on DoubleWidth basic properties and inits.

stephentyrone:
NFC 🤞🏻

### DIFF
--- a/Sources/_TestSupport/DoubleWidth.swift
+++ b/Sources/_TestSupport/DoubleWidth.swift
@@ -48,18 +48,29 @@ public struct DoubleWidth<Base : FixedWidthInteger> {
   public typealias Low = Base.Magnitude
 
   internal var _storage: (low: Low, high: High)
+}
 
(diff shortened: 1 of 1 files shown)


stephentyrone:
@swift-ci please test

Sources/_TestSupport/DoubleWidth.swift
```diff
@@ -79,10 +90,18 @@ public struct DoubleWidth<Base : FixedWidthInteger> {
   //
   // For that reason, we'll include an internal initializer that takes two
   // separate arguments.
+  
+  /// Creates a new instance from the given tuple of high and low parts.
```
xwu, line 94: Technically, no longer takes a tuple since it’s splatted. Also internal (at least at present).

stephentyrone: Doc comments are perfectly useful on internal API =)

PR #213, opened 2021-11-17
add arm64 support for M1 Mac

DanboDuan:
it will fail if it is built on M1 Mac, so add arm64 for M1 Mac


cmake/modules/SwiftSupport.cmake
```diff
@@ -17,8 +17,12 @@ See https://swift.org/LICENSE.txt for license information
 function(get_swift_host_arch result_var_name)
   if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
     set("${result_var_name}" "x86_64" PARENT_SCOPE)
-  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
-    set("${result_var_name}" "aarch64" PARENT_SCOPE)
+  elseif ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "AArch64|aarch64|arm64")
+    if(CMAKE_SYSTEM_NAME MATCHES Darwin)
```
stephentyrone, line 21: I am not a CMake expert; should `CMAKE_SYSTEM_NAME` be string-substituted here (like `${CMAKE_SYSTEM_PROCESSOR}` is?

stephentyrone:
@compnerd and/or @gottesmm, can you review this, please?

cmake/modules/SwiftSupport.cmake
DanboDuan: this cmake code is just copied from swift-atomics/cmake/modules/SwiftSupport.cmake. 
or I can update it to `STREQUAL`

stephentyrone: Ok. I'll take this as-is, then, and if Saleem or Michael have any opinions on how it could be better we'll do that separately.

stephentyrone:
@swift-ci test

stephentyrone (approved):

cmake/modules/SwiftSupport.cmake
compnerd: This matches the other projects, and is fine.  It avoids some of the noise by not dereferencing which is common.

PR #224, opened 2022-04-06
Minor documentation formatting cleanup.

stephentyrone:
Xcode changed how it parses doc comments like

> Header
> -
> - bulleted list

so that now they simply get dropped on the floor. Remove this pattern from our doc comments.


stephentyrone:
@swift-ci test

PR #225, opened 2022-04-06
Remove conformance of Complex to Differentiable

stephentyrone:
The conformance was breaking CI testing, because the CI bots use a macOS version that predates availability of the _Differentiation module, and the compiler-generated differentation operations have neither availability attributes nor checks, so any reference to those operations simply crashes in dyld because the library is not present. We can add availability, but that's not really accurate either, because these operations are not actually binary-stable, so while we wouldn't crash in dyld, we also might not work correctly in the future. We could mark them unavailable on Apple platforms, but that's also not quite right (and further means that they're not going to get a lot of attention). The best solution seems to be to remove the conformance (which was never source-stable).


lorentey (dismissed):
Looks good!

Differentiation has not yet gone through Swift Evolution, and we cannot rely on its continued availability / source stability. Its current presence in this package carries considerable risk to source stability -- any future Swift release may break this code, rendering package releases that include it unusable, whether or not clients call these APIs.

If there are any existing clients, the best option for them is to maintain these additions in a standalone package with different stability expectations.

stephentyrone:
@swift-ci test

stephentyrone:
@compnerd This will definitely break the cmake build as is, can you spare a little time to help me fix it?

stephentyrone:
(In particular, the _NumericsShims target will fail because it no longer has an `include` subdirectory for the header.)

compnerd:
Ah, interesting; where is the header supposed to come from now?

stephentyrone:
It's just `_NumericShims/_NumericShims.h` now.

compnerd:
Ah, the CMake change is easy:

```
diff --git a/Sources/_NumericsShims/CMakeLists.txt b/Sources/_NumericsShims/CMakeLists.txt
index 92ed523..56bef5b 100644
--- a/Sources/_NumericsShims/CMakeLists.txt
+++ b/Sources/_NumericsShims/CMakeLists.txt
@@ -9,6 +9,6 @@ See https://swift.org/LICENSE.txt for license information
 
 add_library(_NumericsShims INTERFACE)
 target_include_directories(_NumericsShims INTERFACE
-  include)
+  ${CMAKE_CURRENT_SOURCE_DIR})
 
 set_property(GLOBAL APPEND PROPERTY SWIFT_NUMERICS_EXPORTS _NumericsShims)
```

This still will break on Windows due to the `m` linkage.  We need to conditionalize that.

stephentyrone:
> This still will break on Windows due to the m linkage. We need to conditionalize that.

Yep. What should it be there?

compnerd:
Correct; just like on macOS, there is no link necessary.

stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

compnerd:
Hmm, I think that the linking approach would work, but missed the CMake change :-(.

PR #228, opened 2022-04-27
Sync latest package changes with `Quaternions` branch

markuswntr:
Syncs the latest changes from `main` with the `Quaternions` branch (commits up to and including https://github.com/apple/swift-numerics/pull/228/commits/964e62c66971a40b807130bc6d731e4154e1f4c2), refines the quaternion module structure (sync with https://github.com/apple/swift-numerics/pull/219, commit: https://github.com/apple/swift-numerics/pull/228/commits/f56d2cb0518c65151c9c4314fa13dfc5a061f2d2) and refines documentation (sync with https://github.com/apple/swift-numerics/pull/224, commit: https://github.com/apple/swift-numerics/pull/228/commits/4f973441cd3bdbf1740b5e6eeab9a390594ae6e3).

PR #229, opened 2022-04-27
Remove conformance to `AlgebraicField` on quaternions

markuswntr:
As of now, `Quaternion` conforms to the `AlgebraicField` protocol defined in `RealModule`. However quaternions are a non-commutative, associative algebra over the real numbers and thus, this is incorrect. Their structure is similiar to a field, but given that their multiplication is non-commutative they form a division algebra (division ring). This has no implications on the implementation of `AlgebraicField` on `Quaternion`, but contradicts the mathematical definition – and the definition in the `AlgebraicField` documentation noting that:

```
... a field is a commutative group under its addition, the non-zero elements of the field form a commutative group under its multiplication, and the distributive law holds.
```

I would like to propose to remove the conformance to `AlgebraicField` and replace it with `SignedNumeric`, but to keep the implementation of the division operators and `reciprocoal`, as these are well defined for quaternions.

---

This PR depends on https://github.com/apple/swift-numerics/pull/228 being merged first, and so is marked as a Draft for now.


markuswntr:
Its worth noting, that we could add another protocol (a `DivisionRing`, or similar in naming) that extends `SignedNumeric` by division operators and `reciprocal` – effectively, what is now defined as an `AlgebraicField` – and then refine `AlgebraicField` on top of the new division algebra protocol by guaranteeing commutativity. Quaternions would then conform to `DivisionRing` while Real and Complex numbers would conform to `AlgebraicField`.
I don't think a `DivisionRing` protocol is of much use, but I am happy to draft an example an to bring this question over to the Swift Forums, if you feel it is worth it.

stephentyrone (approved):

PR #230, opened 2022-05-19
Conform Quaternion to Elementary Functions

markuswntr:
This extends https://github.com/apple/swift-numerics/pull/206 and adds the remaining quaternionic elementary functions; making `Quaternion` fully conform to `ElementaryFunctions`. 🎉
As you might expect, most of the functions draw inspiration from the implementation details of the corresponding elementary function on complex numbers. Nonetheless, I greatly appreciate every comment, feedback or suggestions!

----

This PR looks a bit like a mess right now as it builds upon the main->Quaternions sync PR https://github.com/apple/swift-numerics/pull/228. This had to be done, as it requires the new `Augmented.sum(_:,_:)` that is only available on main. Once https://github.com/apple/swift-numerics/pull/228 is merged, this PR will contain only the changes necessary to provide `ElementaryFunction` conformance (starting with 3ebee8f5f1c501de6e8703fae6d20fb24aa92e89). 
Meanwhile, here is a direkt link to the file declaring conformance and that contains all the relevant changes: [Quaternion+ElementaryFunctions.swift](https://github.com/markuswntr/swift-numerics/blob/442d4a75dd023eb59dc30bdcfe867f081b5d6280/Sources/QuaternionModule/Quaternion%2BElementaryFunctions.swift)


markuswntr:
The implementation is still functional, but I am trying to improve the error bounds on `log` and maybe add an executable to validate the worst-case error (much like the ComplexLog tests), and so I would like to leave it as a draft for now.

PR #235, opened 2022-09-15
Fix edge-cases for Complex.pow(.zero, <Int>), and Real.pow(zero, zero)

stephentyrone:
Previously the complex pow unconditionally returned `.zero`, but they should actually return one of `.infinity`, `.zero`, or `.one` depending on whether the exponent is negative, positive, or zero, matching the behavior of real types.

Similarly, the real pow didn't quite match IEEE 754 powr for the (0, 0) case (should be .nan, was 0).


stephentyrone:
@swift-ci  test

stephentyrone:
@swift-ci test

Issue #242, opened 2023-01-18
[BigInt] Using tests from “Violet - Python VM written in Swift”


LiarPrincess:
LATER…

LiarPrincess:
[3.2.2023] EDIT to main issue:
- [[BigInt tests] 💀 Init from float 754](https://github.com/apple/swift-numerics/pull/258)
- [[BigInt tests] ✅🐰 Codable](https://github.com/apple/swift-numerics/pull/259)
- many changes in performance tests

LiarPrincess:
Things to do in `Helpers` (but I'm not going to update 15 PRs, so we will do it later):
- rename `BigIntPrototype.compare` -> `BigIntPrototype.compareMagnitude`
- rewrite `BigIntPrototype.create` - shifts 10x faster than multiplication:

  ```Swift
  internal static func create<T: FixedWidthInteger>(
    isPositive: Bool,
    magnitude: [T]
  ) -> BigInt {
    assert(!T.isSigned)
    var result = BigInt()

    for (index, word) in magnitude.enumerated() {
      var bits = BigInt(word)
      bits <<= index * T.bitWidth
      result |= bits
    }

    if !isPositive {
      result.negate()
    }

    return result
  }
  ```

Btw. I updated the Violet implementation. Now it should be much faster (see the performance ticket).

LiarPrincess:
I wrote [Violet XsProMax](https://github.com/LiarPrincess/Violet-BigInt-XsProMax) which is a [Violet](https://github.com/LiarPrincess/Violet/tree/swift-numerics) implementation with following changes:
- no small inlined integer (`Smi`) - magnitude is always stored on the heap
- no restrictions on the size - `isNegative` is stored in-line (and not on the heap like in Violet); `count` and `capacity` are on the heap because I don't want to stray too much from `Violet`.

Which gives us:

```Swift
struct BigInt {
  struct Header {
    var count: UInt32
    var capacity: UInt32
  }

  var flags: UInt8
  var buffer: ManagedBufferPointer<Header, UInt>
}
```

I updated the performance PR with the results.

mgriebling:
The issue below appears fixed in my latest updates (see pull request # 261)

``` swift
// -9223372036854775808 = Int64.min, obviously '-Int64.min' overflows
let int: Int64 = -9223372036854775808
let big = BigInt(int)
XCTAssertEqual(-big, big * -1)
```

mgriebling:
[...]

PR #244, opened 2023-01-18
[BigInt tests] ✅ String tests

LiarPrincess:

---

All pass.

## String parsing

I think that `String` parsing performance can be improved. 4000 cases with `BigInts` up to 1280 bits take ~40s on my laptop (2014 rMBP -> mac 11.7 (Big Sur), Xcode 13.2.1, Intel), Violet does this in 1.5s (there is a separate performance PR).

Violet secret:

> Instead of using a single `BigInt` and multiplying it by `radix`,
> we will group scalars into words-sized chunks.
> Then we will raise those chunks to appropriate power and add together.
>
> For example:
> `1_2345_6789 = (1 * 10^8) + (2345 * 10^4) + (6789 * 10^0)`
>
> So, we are doing most of our calculations in fast `Word`,
> and then we switch to slow `BigInt` for a few final operations.

Implemented [here](https://github.com/LiarPrincess/Violet/blob/fe662b422346b5b0a4ac2b86333adad59ff434a6/Sources/BigInt/BigInt%2BString.swift#L82). Should I create an issue for this?


mgriebling:
Noticed poor performance in the BigInt initializer from a string as well.
See pull request # 261.

PR #250, opened 2023-01-18
[BigInt tests] 💀 Binary `*`, `/` and `%`

LiarPrincess:

---

## 💀 Crash

```Swift
func test_div_crash() {
  let lhs = BigInt("-9223372036854775808")!
  let rhs = BigInt("-1")!
  _ = lhs / rhs // Overflow
}
```

## ❌ Failures

```Swift
func test_div_sign() {
  // positive / negative = negative
  var lhs = BigInt("18446744073709551615")!
  var rhs = BigInt("-1")!
  var expected = BigInt("-18446744073709551615")!
  XCTAssertEqual(lhs / rhs, expected)

  // negative / positive = negative
  lhs = BigInt("-340282366920938463481821351505477763074")!
  rhs = BigInt("18446744073709551629")!
  expected = BigInt("-18446744073709551604")!
  XCTAssertEqual(lhs / rhs, expected)
}
```


LiarPrincess:
I added the test for rounding toward 0: `|quotient * rhs| <= |lhs|`, but the current implementation of `BigInt.magnitude` does not work, so this test will fail but not because of `div`.

PR #256, opened 2023-01-18
[BigInt tests][No merge] 🐰 How NOT to write performance tests

LiarPrincess:

---

=== DO NOT MERGE! Discussion only. ===

## 🐰 Discussion

I don't think that those tests are good or anything. They are very naïve and brute-force. This is perfectly 'ok' for an initial implementation, but later we should go back to this topic and choose the correct measure points (which `BigInt` sizes/ranges are the most important etc.).

It may also be a good idea to move all of the performance tests to a separate test target.

You can use [Violet](https://github.com/LiarPrincess/Violet) as a baseline measure, thought Violet was heavily optimized for `Int32` range (see [documentation](https://github.com/LiarPrincess/Violet/blob/main/Documentation/BigInt.md)). I could easily write something faster, but this is not my use-case.

## String parsing

That said, I have to say that I find the `String` parsing performance quite atrocious. 4000 cases with `BigInts` up to 1280 bits take ~40s on my laptop, Violet does this in 1.5s.

Violet secret:

> Instead of using a single `BigInt` and multiplying it by `radix`,
> we will group scalars into words-sized chunks.
> Then we will raise those chunks to appropriate power and add together.
>
> For example:
> `1_2345_6789 = (1 * 10^8) + (2345 * 10^4) + (6789 * 10^0)`
>
> So, we are doing most of our calculations in fast `Word`,
> and then we switch to slow `BigInt` for a few final operations.

Implemented [here](https://github.com/LiarPrincess/Violet/blob/fe662b422346b5b0a4ac2b86333adad59ff434a6/Sources/BigInt/BigInt%2BString.swift#L82). Should I create an issue for this?

## Mac

I have 2014 rMBP -> mac 11.7 (Big Sur), Xcode 13.2.1, Intel. I assume that 9 years old machines are not exactly a priority for 🍎. Executing all of those tests take ~30 min (serial execution, `DEBUG`), and I can't be bothered to re-run them properly since it will throttle after a few minutes anyway…

## Linux

Not tested:

```swift
PerformanceTests.generated.swift:27:23: error: cannot find type 'XCTMetric' in scope
private let metrics: [XCTMetric] = [XCTClockMetric()] // XCTMemoryMetric()?
                      ^~~~~~~~~

PerformanceTests.generated.swift:28:23: error: cannot find 'XCTMeasureOptions' in scope
private let options = XCTMeasureOptions.default
                      ^~~~~~~~~~~~~~~~~
```


LiarPrincess:
Update 20.3.2023:
- Added [Violet XsProMax](https://github.com/LiarPrincess/Violet-BigInt-XsProMax) to the post above

PR #261, opened 2023-04-18
Added improved LosslessStringConvertible init() with "^=" bug fix.

mgriebling:
Added improved LosslessStringConvertible init() with order of magnitude faster performance.
Removed unnecessary _digits and _digitRadix.
Fixed a copy/paste problem in "^=" function. (see Issue # 260)
Added test cases for string conversion performance and logical operation tests.

Added a fix for division/remainder having incorrect signs.
Also fixed a corner-case issue where negated integers weren't.
Also added test cases for the division sign.


mgriebling:
Appreciate the feedback.  Glad to know these times can be improved even more.
I am a bit surprised by lack of support for this from Apple and that so many issues are unresolved.
I do have my own version of BigInt (who doesn't 😉) and will continue on that.
I was even original and called it Integer instead of BigInt.  

Hopefully, one day, Apple will see the light and realize that a log of people want better math support on the Mac and iDevices.  For instance, why does no one have an "ipow" function yet.  Or why don't the binary integers have x.squared() support which is generally could be twice as fast as multiplying x * x.

Good work on Violet, BTW.  The project looks very impressive.

PR #265, opened 2023-07-17
Various documentation improvements.

stephentyrone:
Added some docc catalogs and articles, fixed a bunch of minor errors, all that good stuff. More work needed.

### DIFF
--- a/Sources/ComplexModule/Documentation.docc/ComplexModule.md
+++ b/Sources/ComplexModule/Documentation.docc/ComplexModule.md
@@ -71,4 +71,4 @@ let b = 2*a
 `b` ought to have type `Double`, but if we did have a Complex-by-Real `*` 
 operation, `2*a` would either be ambiguous (if there were no type context),
 or be inferred to have type `Complex<Double>` (if the expression appeared
-in the context of an extension defined on `Compex`).
+in the context of an extension defined on `Complex`).
--- a/Sources/ComplexModule/Documentation.docc/ComplexModule.md
+++ b/Sources/ComplexModule/Documentation.docc/ComplexModule.md
@@ -7,6 +7,7 @@ Types and operations for working with complex numbers.
 The `Complex` type is generic over an associated `RealType`; complex numbers
 are represented as two `RealType` values, the real and imaginary parts of the
 number.
+
 ```
 let z = Complex<Double>(1, 2)
 let re = z.real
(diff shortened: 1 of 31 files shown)


Sources/ComplexModule/Documentation.docc/ComplexModule.md
```diff
@@ -0,0 +1,63 @@
+# ``ComplexModule``
+
+Types and operations for working with complex numbers.
+
+## Representation
+
+The `Complex` type is generic over an associated `RealType`; complex numbers
+are represented as two `RealType` values, the real and imaginary parts of the
+number.
+```
```
xwu, line 10: For broader compatibility with various Markdown viewer implementations, would it be possible to add an extra line before and after such code blocks?

Sources/ComplexModule/Documentation.docc/ComplexModule.md
```diff
@@ -0,0 +1,63 @@
...
+double complex w = r + z;       // 1 + 2i
+```
+The `Complex` type does not provide such mixed operators. There are two
+reasons for this choice. First, Swift generally avoids mixed-type
+arithmetic, period. Second, mixed-type arithmetic operators lead to
+undesirable behavior in common expressions when combined with literal
+type inference. Consider the following example:
+```swift
+let a: Double = 1
+let b = 2*a
```
markuswntr, line 48: I think it is easier to follow along the arithmetic examples (especially the next one), if this unsupported heterogeneous operator example would re-use the example from C above (and below) – I may be wrong.

```suggestion
import ComplexModule
let r: Double = 1.0
let z = Complex<Double>(0, 2)
let w = r + z // ❌
```

stephentyrone: The point is to provide an example that would break _if we did have a heterogeneous operator_.

markuswntr: True, my example is misleading. Does it make sense to adjust the other two examples, and make them use multiplication over addition?

stephentyrone: I tried to rearrange stuff to clarify it a little bit, might revisit later.

markuswntr: LGTM!

Sources/ComplexModule/Scale.swift
```diff
@@ -20,35 +20,22 @@
...
-    // Although these two interpretations are _mathematically_ equivalent,
-    // they will generate different representations of the point at
-    // infinity in general. For example, suppose `self` is represented by
-    // `(infinity, 0)`. Then `self * Complex(1)` would evaluate as
-    // `(1*infinity - 0*0, 0*infinity + 1*0) = (infinity, nan)`, but
-    // the vector space interpretation produces `(infinity, 0)`. This does
-    // not matter much, because these are two representations of the same
-    // semantic value, but note that one requires four multiplies and two
-    // additions, while the one we use requires only two real multiplications.
+  /// `self` multiplied by the real value `a`.
```
natecook1000, line 25:
```suggestion
  /// Returns the result of multiplying this complex number by the real value `a`.
```

Sources/ComplexModule/Scale.swift
```diff
@@ -20,35 +20,22 @@
...
+  /// Equivalent to `self * Complex(a)`, but may be computed more efficiently.
+  @inlinable @inline(__always)
+  public func multiplied(by a: RealType) -> Complex {
     Complex(x*a, y*a)
   }
   
-  /// `self` unscaled by `a`.
-  @usableFromInline @_transparent
-  internal func divided(by a: RealType) -> Complex {
+  /// `self` divided by the real value `a`.
```
natecook1000, line 33:
```suggestion
  /// Returns the result of dividing this complex number by the real value `a`.
```

Sources/IntegerUtilities/DivideWithRounding.swift
```diff
@@ -12,26 +12,29 @@
 extension BinaryInteger {
   /// `self` divided by `other`, rounding the result according to `rule`.
```
natecook1000, line 13:
```suggestion
  /// Returns the result of dividing this value by `other`, rounding the result 
  /// according to `rule`.
```

or (preferred):

```suggestion
  /// Returns the result of dividing this value by the given value, rounding
  /// the result according to specified rule.
```

natecook1000 (commented):
I love it! Added a few notes but this is overall a huge improvement 👏🏻👏🏻👏🏻

PR #271, opened 2023-09-28
Replace `some BinaryInteger` with an explicit generic parameter.

stephentyrone:
For compatibility with pre-Swift-5.7 language versions.


stephentyrone:
@swift-ci test

Issue #272, opened 2023-10-11
Crash in `DoubleWidth`


LiarPrincess:
(See below)

LiarPrincess:
And this:
```Swift
import XCTest
import _TestSupport

class xxx: XCTestCase {

  typealias UInt128 = DoubleWidth<UInt64>
  typealias Word = UInt64

  func test_div_2_by_2() {
    self.div([0xea8a9116b7af33b7, 0x3d9d6779ddd22ca3], [0xc3673efc7f1f37cc, 0x312f661057d0ba94], [               0x1], [0x2723521a388ffbeb,  0xc6e01698601720f])
  }

  private func div(
    _ lhsWords: [Word],
    _ rhsWords: [Word],
    _ quotientWords: [Word],
    _ remainderWords: [Word],
    file: StaticString = #file,
    line: UInt = #line
  ) {
    let lhs = self.create(lhsWords)
    let rhs = self.create(rhsWords)
    let quotient = self.create(quotientWords)
    let remainder = self.create(remainderWords)

    // https://www.wolframalpha.com/input?i=311758830729407788314878278112166161571+%2F+259735543268722398904715765931073125012
    print(lhs) // 311758830729407788314878278112166161571
    print(rhs) // 259735543268722398904715765931073125012
    print(quotient) // 1
    print(remainder) // 52023287460685389410162512181093036559

    let qr = lhs.quotientAndRemainder(dividingBy: rhs)
    print(qr.quotient) // 5816654726380801538
    print(qr.remainder) // 108746657730973513852143929583839715195
    XCTAssertEqual(qr.quotient, quotient, "quotientAndRemainder.quotient", file: file, line: line)
    XCTAssertEqual(qr.remainder, remainder, "quotientAndRemainder.remainder", file: file, line: line)
  }

  // MARK: - Create

  private func create(_ words: [Word]) -> UInt128 {
    switch words.count {
    case 1: return UInt128((0, words[0]))
    case 2: return UInt128((words[0], words[1]))
    default: fatalError("Unknown UInt128 input: \(words)")
    }
  }
}
```

It will not crash (I commented the assertions inside `DoubleWidth`), but the result seems to not be correct.
But, again, I will provide more details a little bit later.

LiarPrincess:
(Found it!)


## History

[...]

PR #273, opened 2023-10-26
[_TestSupport] Fix DoubleWidth think-o that trips an assert

xwu:
If `lhs` and `rhs` have the same number of leading zero bits (and we know `rhs < lhs`), then we can't call `_divide` to divide a triple-width magnitude by a double-width magnitude (this trips the assert on line 634), but rather we know that the quotient is one. Additionally, we must use `>>` instead of `&>>` when computing the high word because in this scenario the leading zero bits may be zero.

Since it turns out there's more than one calling site for `_divide` that doesn't ensure the invariant asserted on line 634, we should actually deal with this in one place by replacing the assertion with a `guard` that handles the special case.

Additionally, fix a third think-o related to full-width multiplication where a carry isn't accounted for, causing unintended overflow with certain inputs.

This type now gives correct output verified for addition/subtraction/multiplication/nonzero-division-with-overflow for every pair of 16-bit unsigned integers using `DoubleWidth<DoubleWidth<UInt4>>` (with a custom `UInt4` implementation).

Resolves https://github.com/apple/swift-numerics/issues/272

### DIFF
--- a/Sources/_TestSupport/DoubleWidth.swift
+++ b/Sources/_TestSupport/DoubleWidth.swift
@@ -735,11 +735,12 @@ extension DoubleWidth : UnsignedInteger where Base : UnsignedInteger {
 
     // Left shift both rhs and lhs, then divide and right shift the remainder.
     let shift = rhs.leadingZeroBitCount
+    let high = (lhs >> (Magnitude.bitWidth &- shift)).low
     let rhs = rhs &<< shift
-    let high = (lhs &>> (Magnitude.bitWidth &- shift)).low
     let lhs = lhs &<< shift
-    let (quotient, remainder) =
-      Magnitude._divide((high, lhs.high, lhs.low), by: rhs)
+    let (quotient, remainder) = high == (0 as Low)
+      ? (1, lhs &- rhs)
+      : Magnitude._divide((high, lhs.high, lhs.low), by: rhs)
     return (Magnitude(0, quotient), remainder &>> shift)
   }
 }
--- a/Sources/_TestSupport/DoubleWidth.swift
+++ b/Sources/_TestSupport/DoubleWidth.swift
@@ -363,8 +363,10 @@ extension DoubleWidth : FixedWidthInteger {
         
     let low =
       DoubleWidth<Low>(mid1.partial, a.partial)
+    let (sum_, overflow_) =
+      mid1.carry.addingReportingOverflow(mid2.partial)
     let high =
-      DoubleWidth(High(mid2.carry + d.carry), mid1.carry + mid2.partial)
+      DoubleWidth(High(mid2.carry + d.carry + (overflow_ ? 1 : 0)), sum_)
         
     if isNegative {
       let (lowComplement, overflow) = (~low).addingReportingOverflow(1)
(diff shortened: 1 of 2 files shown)


LiarPrincess:
I updated dem issue with more of dem details:
- merged both comments
- 1 new test

Current status after this PR:

| Test   | Status                             | Comment                                        |
| ------ | ---------------------------------- | ---------------------------------------------- |
| test_a | 🟢 Works                            | Test 1 from the original issue.                |
| test_b | 🔴 Incorrect quotient and remainder | Test 2 from the original issue. Still failing. |
| test_c | 🔴 Assertion failure                | New test.        |

xwu:
@LiarPrincess Thank you for providing more detail in the bug report regarding related edge cases!

xwu:
@swift-ci test

LiarPrincess:
Aded `test_d` and `test_e` to the issue.

Current status:

| Test   | Status                             | Comment                                        |
| ------ | ---------------------------------- | ---------------------------------------------- |
| test_a | 🟢 Works | |
| test_b | 🟢 Works | |
| test_c | 🟢 Works | |
| test_d | 🔴 Assertion failure | New `UInt256` test. |
| test_e | 🔴 Assertion failure | New `UInt256` test. |

xwu:
@LiarPrincess, have you provided all of the failing tests in the bug report? It would be good to see all of them in one go instead of iteratively expanding the bug.

LiarPrincess:
In the issue I mentioned:

> Those tests come from [Oh-my-decimal](https://github.com/LiarPrincess/Oh-my-decimal):
> - [UInt128Tests](https://github.com/LiarPrincess/Oh-my-decimal/blob/mr-darcy/Tests/DecimalTests/Generated/GeneratedUInt128Tests.swift) and [UInt256Tests](https://github.com/LiarPrincess/Oh-my-decimal/blob/mr-darcy/Tests/DecimalTests/Generated/GeneratedUInt256Tests.swift) - you may need to modify them to compile with `DoubleWidth` - see "Tests" section below.
> - [Python script to generate tests](https://github.com/LiarPrincess/Oh-my-decimal/blob/mr-darcy/Scripts/generate-tests/wide_uint_tests.py) - the quality of the test cases is not that great - just some random numbers smashed together -> they will not excise edge cases for `min/max` etc. Only use unsigned types, so you have check any signed types yourself.

To run those tests on `DoubleWidth` you need to:
1. add this inside every class:

  ```Swift
  typealias UInt128 = DoubleWidth<UInt64>
  typealias UInt256 = DoubleWidth<UInt128>
  typealias Word = UInt64
  ```

2. customize the `create` method - copy from the test classes that I provided in the issue
3. comment the "multiply by half" section - compiler error, just comment the red squigglies.

xwu:
@LiarPrincess, that's a lot of tests! Which ones fail?

LiarPrincess:
You can just run them: copy the code from the gist and run it. They should finish execution in ~1 second. If there is a failure then Xcode will point you to an exact line. If there is a crash then it will automatically jump into a debugger.In the issue I added 2 examples: test_d and test_e. I hope that resolving them will make all tests pass, but since I have not debugged your code it may be possible that there is some other bug hiding in there.Sent from my iPhoneOn 31 Oct 2023, at 14:42, Xiaodi Wu ***@***.***> wrote:﻿
@LiarPrincess, that's a lot of tests! Which ones fail?

xwu:
@swift-ci test

xwu:
@swift-ci test

LiarPrincess:
| Test   | Status                             |
| ------ | ---------------------------------- |
| `UInt128` tests from gist | 🟢 Works | 
| `UInt256` tests from gist | 🟢 Works | 

I also tested `DoubleWidth` in some of my projects (including the decimal thingie), and it seems to be working. Note that I only use unsigned types (`UInt128/UInt256`), and I do not test edge cases (`UInt128.max` = all bits 1 etc.), so some paths are still not tested. In terms of performance `DoubleWidth` seems to be 25-35% slower than direct implementation.

Anyway, I conform that this PR fully solves [#272](https://github.com/apple/swift-numerics/issues/272).

xwu:
Great, cleaned up the commit history; this is ready to go. Will update main apple/swift repo soon.

xwu:
@swift-ci test

stephentyrone:
@swift-ci test

stephentyrone:
Linux test failure is a misconfiguration of CI, which I've notified Mishal of. We'll merge this as-is.

Issue #275, opened 2023-12-08
Variadic GCD and LCM


LePips:
While #10 exists, I will use this as a separate tracker for these two algorithms and their variadic structure.

The existing `gcd` can be made easily to take a variadic amount of values, along with some checks. `lcm` can also be implemented using the `gcd` method, also variadic.

Possible question: do we want this variadic structure? It does allow 0 and 1 parameters, which are easy cases to account for. It just seems funky that you can put `gcd()` and `lcm()` 🤷

PR #276, opened 2023-12-08
Variadic GCD and LCM

LePips:
- Tracked by: #275

Implemented variadic GCD and LCM. LCM uses the GCD method with accounting for `gcd == 0`.

TODO:
- [ ] Finish documentation
- [ ] Test cases
- [ ] Account for special cases with `Int.min`

### DIFF
--- a/Sources/IntegerUtilities/LCM.swift
+++ b/Sources/IntegerUtilities/LCM.swift
@@ -26,6 +26,9 @@ public func lcm<T: BinaryInteger>(_ n: T...) -> T {
 
 @inlinable
 internal func _lcm<T: BinaryInteger>(_ a: T, _ b: T) -> T {
+    
+    // Using the gcd algorithm with accounting
+    // for possible overflow of x*y
     let x = T(a.magnitude)
     let y = T(b.magnitude)
     
--- a/Sources/IntegerUtilities/LCM.swift
+++ b/Sources/IntegerUtilities/LCM.swift
@@ -1,0 +1,31 @@
+//===--- LCM.swift --------------------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
+// Copyright (c) 2021 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+/// The least common multiple of a given list of values.
+///
+/// If no values are provided, the result is zero.
+///
+/// TODO
+@inlinable
+public func lcm<T: BinaryInteger>(_ n: T...) -> T {
+    guard let first = n.first else { return 0 }
+    guard n.count > 1 else { return first}
+    
+    return n.reduce(first, _lcm)
+}
+
+@inlinable
+internal func _lcm<T: BinaryInteger>(_ a: T, _ b: T) -> T {
+    let x = T(a.magnitude)
+    let y = T(b.magnitude)
+    
+    return x * (y / gcd(x, y))
+}
--- a/Sources/IntegerUtilities/LCM.swift
+++ b/Sources/IntegerUtilities/LCM.swift
@@ -14,6 +14,8 @@
 /// If no values are provided, the result is zero.
 ///
 /// TODO
+///
+/// [wiki]: https://en.wikipedia.org/wiki/Least_common_multiple
 @inlinable
 public func lcm<T: BinaryInteger>(_ n: T...) -> T {
     guard let first = n.first else { return 0 }
@@ -27,5 +29,9 @@ internal func _lcm<T: BinaryInteger>(_ a: T, _ b: T) -> T {
     let x = T(a.magnitude)
     let y = T(b.magnitude)
     
-    return x * (y / gcd(x, y))
+    let z = gcd(x, y)
+    
+    guard z != 0 else { return 0 }
+    
+    return x * (y / z)
 }
--- a/Sources/IntegerUtilities/GCD.swift
+++ b/Sources/IntegerUtilities/GCD.swift
@@ -14,6 +14,8 @@
 /// If no values are provided or all values are zero, the result is zero.
 /// If one input is zero, the result is the absolute value of the other input.
 ///
+/// TODO
+///
 /// The result must be representable within its type. In particular, the gcd
 /// of a signed, fixed-width integer type's minimum with itself (or zero)
 /// cannot be represented, and results in a trap.
--- a/Sources/IntegerUtilities/GCD.swift
+++ b/Sources/IntegerUtilities/GCD.swift
@@ -9,10 +9,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// The greatest common divisor of `a` and `b`.
+/// The greatest common divisor of a given list of values.
 ///
-/// If both inputs are zero, the result is zero. If one input is zero, the
-/// result is the absolute value of the other input.
+/// If no values are provided or all values are zero, the result is zero.
+/// If one input is zero, the result is the absolute value of the other input.
 ///
 /// The result must be representable within its type. In particular, the gcd
 /// of a signed, fixed-width integer type's minimum with itself (or zero)
@@ -23,7 +23,15 @@
 ///
 /// [wiki]: https://en.wikipedia.org/wiki/Greatest_common_divisor
 @inlinable
-public func gcd<T: BinaryInteger>(_ a: T, _ b: T) -> T {
+public func gcd<T: BinaryInteger>(_ n: T...) -> T {
+    guard let first = n.first else { return 0 }
+    guard n.count > 1 else { return first }
+    
+    return n.reduce(first, _gcd(_:_:))
+}
+
+@inlinable
+internal func _gcd<T: BinaryInteger>(_ a: T, _ b: T) -> T {
   var x = a.magnitude
   var y = b.magnitude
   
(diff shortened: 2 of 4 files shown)


stephentyrone:
In practice, variadic functions for these are going to be less useful than operations on sequences:
```
@inlinable
public func gcd<T: BinaryInteger>(_ sequence: some Sequence<T>) -> T {
  sequence.reduce(0, gcd(_:_:))
}

@inlinable
public func lcm<T: BinaryInteger>(_ sequence: some Sequence<T>) -> T {
  sequence.reduce(1, lcm(_:_:))
}
```
Rationale: if you have the sequence version, you can just wrap your values in an array. But if all you have is the variadic version and you have a sequence, you're out of luck.

LePips:
I did think about that, however I think that the correct place for the `gcd/lcm` of a `Sequence` is an extension to match other semantics of "what is the X value in this sequence".

Additionally, that can be easily implemented, as you already do.

stephentyrone:
That doesn't argue for making them variadics, though. That would argue for making them an extension on Sequence.

I think I disagree: `gcd(someValues)` reads more naturally to me than `someValues.gcd()`, even if the latter is more typical for Swift. I'm open to discussing this point, however.

Sources/IntegerUtilities/LCM.swift
```diff
@@ -0,0 +1,37 @@
...
+//===----------------------------------------------------------------------===//
+
+/// The least common multiple of a list of values.
+///
+/// TODO
+///
+/// [wiki]: https://en.wikipedia.org/wiki/Least_common_multiple
+@inlinable
+public func lcm<T: BinaryInteger>(_ n: T...) -> T {
+    assert(!n.isEmpty, "lcm must be positive")
```
stephentyrone, line 19: The LCM of the empty set is 1.

LePips:
I see their value as variadics and extensions on sequence - Swift [has the variadic freestanding](https://developer.apple.com/documentation/swift/min(_:_:_:_:)) for `min/max` as well as the `min` on `Sequence`. I can implement them here as well.

LePips:
Do we actually want to mirror Swift's structure of the binary and variadic `min/max` for the freestanding implementations?

ex: have `lcm(_:_)` and `lcm(_:_:_:_:)`

stephentyrone:
The variadic min/max probably don't carry their weight; I don't think that we would add them if we were writing the standard library today, but they date back to the very, very early days of Swift.

Sources/IntegerUtilities/LCM.swift
```diff
@@ -0,0 +1,35 @@
...
+
+/// The least common multiple of a list of values.
+///
+/// TODO
+///
+/// [wiki]: https://en.wikipedia.org/wiki/Least_common_multiple
+@inlinable
+public func lcm<T: BinaryInteger>(_ n: T...) -> T {
+    n.reduce(1, _lcm(_:_:))
+}
```
markuswntr, line 20: I am assuming there aren't a lot of algorithms based around a variadic input and these functions are rather useful in a context where the number of parameter is known at the time of calling these functions: In this case I wonder if it is worth to skip the empty set entirely, since it does look like an error and we instead enforce at least one value.

```suggestion
@inlinable
public func lcm<T: BinaryInteger>(_ a: T, _ n: T...) -> T {
    _lcm(a, n.reduce(1, _lcm(_:_:)))
}
```

The same applies to the variadic `gcd`, leaving the empty set to `Sequence` based implementations for both of these functions instead (which I expect will be used by algorithms that build upon either of the two).

LePips:
I personally do not have a situation where I need to compute the `lcm/gcd` often enough to make a strong argument for anything. The argument for freestanding variadic could be that I already have my arguments, I don't need to put them into an array first.

An extension on `Sequence` is only natural to fit the functional paradigm. Frankly, I would see many people making that extension on their own if this package didn't provide it.

Sources/IntegerUtilities/LCM.swift
LePips: I am fine with this and non-empty enforcement.

LePips:
Apologies, after my caffeine high I have updated the PR description with the proper reasoning and justifications for this proposal which should have been there to begin with.

JimWallace:
What's the reason for not including both variadic and Sequence versions? 

Seems like both might be useful (though I'd agree that the free function that takes a Sequence<BinaryInteger> is what I would have reached for).

stephentyrone:
> What's the reason for not including both variadic and Sequence versions?

It's generally desirable for a library to make choices about API, rather than just throwing in everything you think someone might want. You want to have a consistent and familiar way to perform operations when you can. Given the sequence form, it's trivial for users to get the variadic behavior by just making an array at the call site: `gcd([a, b, c])`, but there's no convenient way to go from the variadic form to an operation on sequences. So the sequence form makes the most sense.

Why provide the binary form, then, since that can also be spelled using the sequence form? Two reasons: first, the overwhelming majority of use is for exactly two values, so it's justifiable to carve out a specific binding for that. Second, you need to have the binary form to implement the sequence API anyway, and simply making it public doesn't add any implementation or maintenance burden.

JimWallace:
That makes sense, but isn’t the Variadic form strictly more useful than the binary form? I’d take Variadic over binary as an API consumer.

This is probably a lot of handwringing over a tiny number of use cases, but if the API surface area is a concern Variadic and Sequence-oriented implementations could be public, and the binary form could be a fileprivate implementation detail.

markuswntr:
> In practice, variadic functions for these are going to be less useful than operations on sequences:

I do agree. A more elaborate algorithm will almost always have to resort to a function taking a sequence rather than a function taking variadic arguments.

> I did think about that, however I think that the correct place for the `gcd/lcm` of a `Sequence` is an extension to match other semantics of "what is the X value in this sequence".

Making `gcd`/`lcm` instance functions via an extension on `Sequence` (e.g. `values.gcd()`) reads very unnatural to me. If they should not be part of the global namespace (which is probably a good thing), I would rather prefer type functions, which also matches what has been done with `ElementaryFunctions` (and others) on `Real`s (i.e. making the function static and calling them like so: `.gcd(values)`) – but this diverges from the freestanding binary `gcd`, so we probably do not want to do this either and just make both freestanding.

> This is probably a lot of handwringing over a tiny number of use cases, but if the API surface area is a concern Variadic and Sequence-oriented implementations could be public, and the binary form could be a fileprivate implementation detail.

I can not think of a use case for a variadic function of gcd/lcm that passes less than 2 arguments. If we were to limit the argument count to >= 2, we could merge the binary function and the variadic function like so (or similar):

```suggestion
public func gcd<T: BinaryInteger>(_ a: T, _ b: T, _ n: T...) -> T {
  if n.isEmpty {
    return _gcd(a, b)
  }
  return n.reduce(_gcd(a, b), _gcd(_:_:))
}
```

This yields us both, a function that takes 2 or any number of arguments. It does not increase the API surface and it's not a source breaking change. Also, it would not be a performance regression for the binary function **if** the optimiser is able to replace it with the underlying binary function in cases where only 2 arguments are given – although I don't expect it is much of a performance hit otherwise.

LePips:
> In practice, variadic functions for these are going to be less useful than operations on sequences:

I do also agree, however I reiterate that variadic parameters are targeted towards an API implementation allowance where a user may want to use _n_ arguments without making an array/collection. It's nothing more than something nice-to-have as a user and I personally don't see a reason to exclude it just because there can also be a sequence version.

As an API design, I find a variadic form robust in usage and description. In fact, the [cpp lcm documentation example](https://en.cppreference.com/w/cpp/numeric/lcm) uses the binary form to make a variadic form. Now obviously just because something is easy to implement and whether it belongs here [has already been said](https://github.com/apple/swift-numerics/pull/169#issuecomment-770975723) and is an ongoing argument in any API design.

> Making gcd/lcm instance functions via an extension on Sequence (e.g. values.gcd()) reads very unnatural to me.

I argue the opposite, but this is very strongly a design choice in functional programming. Swift allows very tacit function composition via type extensions and I am confident in saying that every Swift developer has used the style - either consciously or unknowingly. I could somewhat dive into this into detail, however that's blog post material.

> If they should not be part of the global namespace (which is probably a good thing)

These functions are already namespaced by modules. I don't think that it's a good thing to differ from many other language's math/numerics packages that offer these types of functions as freestanding. I don't see namespace crowding as an issue here and it is consistent with other language packages.

> I can not think of a use case for a variadic function of gcd/lcm that passes less than 2 arguments.

I am personally fine with this for a developer usage reason and tool-autocomplete reason. If a developer is explicitly using it for one argument, I don't see a reason to disallow it - that's just odd. Additionally, when using autocomplete Xcode (at least) will also add a placeholder for the variadic argument - almost implying that it's required but it could be removed by manual deletion. In instances where auto-complete isn't available, then the developer is very much in a context where they know how many variables they would need to evaluate a function with.

I can also see how my argument is contradictory to avoiding the full-variadic argument. However I'm fine with that contradiction, it wouldn't allow something even _more_ weird and autocomplete tools already guide the developer towards a >= 2 argument context.

---

At the same time, I'm not arguing for variadic-exclusivity with a sequence argument equivalent. I can actually see uses of all three styles depending on developer context - and there's very little overhead. Obviously we should be conscious that we are making decisions that would impact all future numerical function implementations and I have left some of my thoughts on the PR description.

Specifically for the sequence version, I would lean more towards these functions having named parameters so that the call site reads more easily. I argue that we see this more in standard library functions (not just to avoid conflicts, but also as an API detail) and that this wouldn't diverge from other language implementations (because I wasn't able to find one that implemented an array/collection version).

So in short, I think we should offer all three and I don't think that it introduces much overhead at all to us or users.

```swift
lcm(a, b, c)
lcm(of: foo)
bar.lcm()
```

PR #280, opened 2024-03-04
Remove unused `LinuxMain.swift` and `WindowsMain.swift`

MaxDesiatov:
These files are no longer needed, as `swift test` enables test discovery by default.


MaxDesiatov:
@swift-ci test

stephentyrone (approved):

PR #282, opened 2024-03-14
Update license headers and copyright notices

markuswntr:
There are a few slightly different versions of license headers distributed along the files of the projects, so I tried to unify them using the existing headers and the template from `swift-collection` (https://github.com/apple/swift-collections/commit/540348f1475110b307e6c59b394c70d1468da229) - and update the year of all copyright notices in the process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 #[[
 This source file is part of the Swift Numerics open source project
 
-Copyright (c) 2019 Apple Inc. and the Swift Numerics project authors
+Copyright (c) 2019 - 2024 Apple Inc. and the Swift project authors
 Licensed under Apache License v2.0 with Runtime Library Exception
 
 See https://swift.org/LICENSE.txt for license information
--- a/Package.swift
+++ b/Package.swift
@@ -3,7 +3,7 @@
 //
 // This source file is part of the Swift Numerics open source project
 //
-// Copyright (c) 2019-2021 Apple Inc. and the Swift Numerics project authors
+// Copyright (c) 2019 - 2024 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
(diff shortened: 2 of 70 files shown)


NevinBR:
Why are you trying to remove the word “Numerics” from the project name in the copyright notice?

markuswntr:
It really is just a copy of the template used in related projects, e.g.:

- [swift-argument-parser](https://github.com/apple/swift-argument-parser/blob/main/Sources/ArgumentParser/Completions/BashCompletionsGenerator.swift#L5)
- [swift-algorithms](https://github.com/apple/swift-algorithms/blob/main/Sources/Algorithms/Chain.swift#L5)
- [swift-collections](https://github.com/apple/swift-collections/blob/main/Sources/BitCollections/BitArray/BitArray.swift#L5)

Neither have the project name in the copyright notice - but still in the first line of the license header.

stephentyrone:
The existing text was discussed with our lawyers chose when we set up the project, which somewhat predates most of the other packages (and hence some of the decisions made around them). Absent a need to change it, we'll keep it as-is.

markuswntr:
> The existing text was discussed with our lawyers chose when we set up the project, which somewhat predates most of the other packages (and hence some of the decisions made around them).

Just to make sure this was not missed, because the PR description could have been clearer: The main goal of this PR was not to change the existing text. As of now, there are a few files in the project that do not have the license header that was chosen when you set up the project, but a slightly different one.
The license header in [Complex+ElementaryFunctions.swift](https://github.com/apple/swift-numerics/blob/main/Sources/ComplexModule/Complex%2BElementaryFunctions.swift), [_TestSupport/Error.swift](https://github.com/apple/swift-numerics/blob/main/Sources/_TestSupport/Error.swift) and others, for example, seem to be a copy from the license header template in [apple/swift](http://github.com/apple/swift), while others contain a copy from the template in related projects.
So I tried to fix the invalid ones and in doing so I chose the version that is used in the related projects I mentioned above.

If this is something that should be addressed, I can reopen the branch and update the files to match the text chosen at the project setup - but I am fine to leave the PR closed if this is not relevant.

stephentyrone:
They appear to all be files that I authored completely (except for one where @NevinBR modified some comments), so I can simply fix that up.

PR #283, opened 2024-03-14
Make `Complex.magnitude` use the `Magnitude` type of the underlying `RealType`

markuswntr:
Currently, the type of `Complex.magnitude` is equal to the underlying `Real` type of the complex number. This is fine for Float[16|32|64] et al, but I believe incorrect, since `Real` does not necessary require `Magnitude` to always be of the same type.

### DIFF
--- a/Sources/ComplexModule/Complex+Numeric.swift
+++ b/Sources/ComplexModule/Complex+Numeric.swift
@@ -49,7 +49,7 @@ extension Complex: Numeric {
   ///
   /// See also `.length` and `.lengthSquared`.
   @_transparent
-  public var magnitude: RealType {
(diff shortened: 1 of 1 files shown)


NevinBR:
`Real` refines `FloatingPoint` which refines `SignedNumeric` with the constraint `where Magnitude == Self`.

See the [declaration of FloatingPoint](https://github.com/apple/swift/blob/main/stdlib/public/core/FloatingPoint.swift#L164-L165) in the standard library.

markuswntr:
Thanks so much for the feedback!

I did not know this is a constraint on `FloatingPoint`  – closing this PR then.

PR #289, opened 2024-05-16
Rework complex division

stephentyrone:
Replaces the rescaling algorithm for Complex division to one inspired by Doug Priest's "Efficient Scaling for Complex Division," with some further tweaks to:
1. allow it to work for arbitrary FloatingPoint types, including Float16
2. get exactly the same rounding behavior as the un-rescaled path, so that z/w = tz/tw when tz and tw are computed exactly.
3. allow future optimizations to hoist a rescaled reciprocal for more speedups.

Unlike Priest, we _do not_ try to avoid spurious overflow in the final computation when the result is very near the overflow boundary but cancellation brings us just inside it. We do not believe that this is a good tradeoff, as complex multiplication overflows in exactly the same way. We will investigate providing opt-in API to avoid this overflow case in a future PR.

### DIFF
--- a/Sources/ComplexModule/Complex+AlgebraicField.swift
+++ b/Sources/ComplexModule/Complex+AlgebraicField.swift
@@ -27,52 +27,95 @@ extension Complex: AlgebraicField {
   }
   
   @_transparent
+  public static func /=(z: inout Complex, w: Complex) {
+    z = z / w
+  }
+  
+  @_transparent
   public static func /(z: Complex, w: Complex) -> Complex {
-    // Try the naive expression z/w = z*conj(w) / |w|^2; if we can compute
-    // this without over/underflow, everything is fine and the result is
-    // correct. If not, we have to rescale and do the computation carefully.
+    // Try the naive expression z/w = z * (conj(w) / |w|^2); if we can
+    // compute this without over/underflow, everything is fine and the
+    // result is correct. If not, we have to rescale and do the
+    // computation carefully (see below).
     let lenSq = w.lengthSquared
     guard lenSq.isNormal else { return rescaledDivide(z, w) }
     return z * (w.conjugate.divided(by: lenSq))
   }
   
-  @_transparent
-  public static func /=(z: inout Complex, w: Complex) {
-    z = z / w
-  }
-  
-  @usableFromInline @_alwaysEmitIntoClient @inline(never)
+  @inline(never)
+  @_specialize(exported: true, where RealType == Float)
+  @_specialize(exported: true, where RealType == Double)
+  @usableFromInline
   internal static func rescaledDivide(_ z: Complex, _ w: Complex) -> Complex {
     if w.isZero { return .infinity }
-    if z.isZero || !w.isFinite { return .zero }
-    // TODO: detect when RealType is Float and just promote to Double, then
-    // use the naive algorithm.
-    let zScale = z.magnitude
-    let wScale = w.magnitude
-    let zNorm = z.divided(by: zScale)
-    let wNorm = w.divided(by: wScale)
-    let r = (zNorm * wNorm.conjugate).divided(by: wNorm.lengthSquared)
-    // At this point, the result is (r * zScale)/wScale computed without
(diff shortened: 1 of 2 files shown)


stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

Sources/ComplexModule/Complex+AlgebraicField.swift
```diff
@@ -27,52 +27,93 @@ extension Complex: AlgebraicField {
...
+    //  does the product as (zw̅ʺ)*1/|wʹ|², while we do zʹ(w̅ʹ/|wʹ|²). We
+    //  prefer our version for three reasons:
+    //
+    //  1. it extracts a little more ILP
+    //  2. it makes it so that we get exactly the same roundings on the
+    //     rescaled divide path as on the fast path, so that z/w = tz/tw
+    //     when tz and tw are computed exactly.
+    //  3. it unlocks a future optimization where we hoist s and
+    //     (w̅ʹ/|wʹ|²) and make divisions all fast-path without perturbing
+    //     rouding.
```
glessard, line 108:
```suggestion
    //     rounding.
```

Sources/ComplexModule/Complex+AlgebraicField.swift
```diff
@@ -27,52 +27,93 @@ extension Complex: AlgebraicField {
...
+    //  which has magnitude like |w|^(-¼).
+    //
+    //  Note that because the scale factor is always a power of the radix,
+    //  the rescaling does not affect rounding, and so this algorithm is scale-
+    //  invariant compared to the mainline `/` implementation, up to the
+    //  underflow boundary.
+    //
+    //  Note that our final assembly of the result is different from Priest;
+    //  he applies s to w twice, instead of once to w and once to z, and
+    //  does the product as (zw̅ʺ)*1/|wʹ|², while we do zʹ(w̅ʹ/|wʹ|²). We
```
glessard, line 99:
```suggestion
    //  does the product as (zw̅ʺ)/|wʹ|², while we do zʹ(w̅ʹ/|wʹ|²). We
```
The `*1` either overcomplicates the expression, or `1/|w'|²` could be between parentheses.

Tests/ComplexTests/ArithmeticTests.swift
```diff
@@ -63,93 +83,101 @@ final class ArithmeticTests: XCTestCase {
...
+              product.real *= q.length
+              product.imaginary *= q.length
+              if checkMultiply(z, w, expected: product, ulps: 16) { XCTFail() }
+              var quotient = Complex(length: p.length, phase: p.phase - q.phase)
+              quotient.real /= q.length
+              quotient.imaginary /= q.length
+              if checkDivide(z, w, expected: quotient, ulps: 16) { XCTFail() }
+            }
+          }
         }
```
glessard, line 168: The indentation of `testPolar()` seems wrong

Tests/ComplexTests/ArithmeticTests.swift
```diff
@@ -191,16 +219,38 @@ final class ArithmeticTests: XCTestCase {
...
     ]
     for test in vectors {
-      if checkDivide(test.a, test.b, expected: test.c, ulps: 0.5) { XCTFail() }
+      if checkDivide(test.a, test.b, expected: test.c, ulps: 1.0) { XCTFail() }
       if checkDivide(test.a, test.c, expected: test.b, ulps: 1.0) { XCTFail() }
       if checkMultiply(test.b, test.c, expected: test.a, ulps: 1.0) { XCTFail() }
     }
   }
-
+
```
glessard, line 227: Nitpicking trailing whitespace…

glessard (approved):
A few nits picked

Sources/ComplexModule/Complex+AlgebraicField.swift
stephentyrone: The reciprocal is load-bearing here.

stephentyrone:
@swift-ci test

PR #293, opened 2024-07-01
Integer midpoint

stephentyrone:
## Overflow-safe integer midpoint with rounding control

A common source of integer and pointer overflows and out-of-bounds accesses across programming languages is midpoint computations for the purposes of collection algorithms. Swift's checked arithmetic and accesses prevent errors, but may still result in spurious crashes when this operation is implemented incorrectly.

Further, rounding-control for integer arithmetic is subtle and easy to get wrong, and further complicates overflow prevention (midpoint _should_ never overflow, no matter how it is rounded, but achieving this can be tricky). For some algorithms it is desirable to round up or down or to even; we simply implement all of the options.

Draft because I am not sold on the free-function spelling `midpoint(a, b)`. It is desirable by symmetry with `min(a, b)` and `max(a, b)`, and because it correctly captures that this is a commutative operation; neither operand is privileged. But Swift generally eschews free functions. This could equally be a static member (`Int.midpoint(a, b)`), or possibly use some other spelling.

### DIFF
--- a/Sources/IntegerUtilities/Midpoint.swift
+++ b/Sources/IntegerUtilities/Midpoint.swift
@@ -1,0 +1,108 @@
+//===--- Midpoint.swift ---------------------------------------*- swift -*-===//
+//
+// This source file is part of the Swift Numerics open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift Numerics project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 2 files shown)


stephentyrone:
@swift-ci test

glessard (approved):

glessard:
@swift-ci please smoke test

glessard:
@swift-ci please test

stephentyrone:
@swift-ci test

PR #294, opened 2024-07-01
Some file headers have the Swift.org text instead of Swift Numerics

stephentyrone:
These are a copy-pasto from introduced when I originally imported approximate equality (which was originally proposed for the standard library).


stephentyrone:
@swift-ci test

PR #295, opened 2024-07-10
Improvements and bugfixes for DoubleWidth

stephentyrone:
These are used only in TestSupport, but it's still nice to get them right. Signed types had a long-standing bug in how overflow was computed for multiplication, and masking shifts would do the wrong thing when the bitwdith was not a power of two and the shift count was negative. I also added implementations of `&*`, `&+`, and `&-`.

### DIFF
--- a/Sources/_TestSupport/DoubleWidth.swift
+++ b/Sources/_TestSupport/DoubleWidth.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift Numerics open source project
 //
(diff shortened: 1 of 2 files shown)


stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test macos

stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

PR #296, opened 2024-07-19
Add new rounding modes: toNearestOr[Down,Up,Zero,Away]

stephentyrone:
These were omitted in the first pass over integer rounding rules, but are generally useful and make good sense to have available. In particular, toNearestOrUp is the natural mode for a lot of fixed-point arithmetic, and frequently has HW support on SIMD units, so it makes good sense to have a name for that. Once you add that, having nearestOrDown also makes sense, and then nearestOrZero should exist by analogy to nearestOrAway. Also beefed up testing and refactored the division rounding code somewhat.

Also renamed `toNearestOrAwayFromZero` to `toNearestOrAway` for symmetry. The original name was never released, but we'll keep it around in deprecated form anyway.

### DIFF
--- a/Sources/IntegerUtilities/DivideWithRounding.swift
+++ b/Sources/IntegerUtilities/DivideWithRounding.swift
@@ -64,44 +64,58 @@ extension BinaryInteger {
       // disagree, we have to adjust q downward and r to match.
       if other.signum() != r.signum() { return q-1 }
       return q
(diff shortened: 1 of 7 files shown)


stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

PR #297, opened 2024-08-15
Make all (IntegerUtilities) rounding rules available to FloatingPoint

stephentyrone:
IntegerUtilities has RoundingRule that is a superset of FloatingPointRoundingRule; many of these are also useful when working with floating-point. This change provides a `rounding(_:)` function similar to the standard libraries `rounded(_:)` that makes all of them available. It has a different name instead of being a shadow because otherwise existing use sites like `rounded(.down)` would become ambiguous.


stephentyrone:
@swift-ci test

PR #310, opened 2025-06-06
Cleaned up GCD and updated GCD tests to Swift Testing

jasonbobier:
By returning T.Magnitude instead of T, gcd(a, b) no longer will produce fatal errors and removes a few special case lines of code. This allows the end user of the api to decide how to handle Signed Fixed Width Integer overflows.

Moved the tests to Swift Testing and updated for new code.

Also renamed the files for clarity and to match the other naming in the IntegerUtilties.


jasonbobier:
Note that this is also prep for an lcm implementation which will be in a separate PR.

jasonbobier:
Ended up adding the lcm to here as well. With fixed width integers, lcm can overflow fairly often. In that case, this code will throw an error that contains the overflow value so that the caller has a chance to use the lcm result in a larger type.

Issue #103, opened 2020-03-12
`SIMD{n}: ElementaryFunctions` conformances


dan-zheng:
`ElementaryFunctions` was added to apple/swift with [conformances for `SIMD` types](https://github.com/stephentyrone/swift/blob/8708c7b5819e45e1f44fafd4e0ad02affb3535ad/stdlib/public/core/MathFunctions.swift.gyb#L128-L172). It seems those conformances were dropped when `ElementaryFunctions` was moved to apple/swift-numerics.

Is it reasonable to re-add `SIMD{n}: ElementaryFunctions` conformances to apple/swift-numerics?

---

Context: https://github.com/tensorflow/swift/issues/404

stephentyrone:
Yup, this is a very reasonable thing to add, I'll take a look shortly.

PR #312, opened 2025-07-21
add ElementaryFunctions conformance to SIMD types

JaapWijnen:
Adresses #103 
There's a few parts that probably require discussion. 

`ElementaryFunctions` requires conformance to `AdditiveArithmetic`
The concrete SIMD2, 4, etc types from the standard library have a conformance to `AdditiveArithmetic` defined in the Differentiation module so this conformance is only visible when `_Differentiation` is imported. 
It's defined [as follows](https://github.com/swiftlang/swift/blob/04726c04e3f112ff74a3d953fefe40d9c761734b/stdlib/public/Differentiation/SIMDDifferentiation.swift.gyb#L26C1-L27C1): `extension SIMD${n}: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}`

In some of our code that would be used with swift toolchains both with and without access to the `_Differentiation` module we'd get around this by using the following:
```swift
#if canImport(_Differentiation)
import _Differentiation
#endif

#if !canImport(_Differentiation)
// add `AdditiveArithmetic` conformance since this is only present in the _Differentiation module which is not present everywhere
extension SIMD2: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
extension SIMD4: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
extension SIMD8: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
extension SIMD16: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
extension SIMD32: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
extension SIMD64: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
#endif
```

My main worry here is that the conformance to `AdditiveArithmetic` in the standard library will conflict with the one provided here. Either the full one or the one that's limited to types conforming to `FloatingPoint`. But as far as I understand this should be ok due to the `@retroactive` attribute. 

Furthermore, technically we could be more generic (here and in the stdlib _Differentiation module) by requiring `Scalar` to conform to just `AdditiveArithmetic` but that would require more boilerplate, now we get the implementation [for free from the standard library](https://github.com/swiftlang/swift/blob/04726c04e3f112ff74a3d953fefe40d9c761734b/stdlib/public/core/SIMDVector.swift#L871). The other solution is more complete however and would look something like: 

```swift
extension SIMD2: @retroactive AdditiveArithmetic where Scalar: AdditiveArithmetic {
    @_transparent
    public static var zero: SIMD2<Scalar> {
        Self()
    }
    
    @_transparent
    public static func + (lhs: SIMD2<Scalar>, rhs: SIMD2<Scalar>) -> SIMD2<Scalar> {
        var v = Self()
        for i in v.indices {
            v[i] = lhs[i] + rhs[i]
        }
        return v
    }
    
    @_transparent
    public static func - (lhs: SIMD2<Scalar>, rhs: SIMD2<Scalar>) -> SIMD2<Scalar> {
        var v = Self()
        for i in v.indices {
            v[i] = lhs[i] - rhs[i]
        }
        return v
    }
}
```

Ideally the `AdditiveArithmetic` conformance would live in the stdlib itself. That way none of the Differentiation code would break and there's no longer the need for this duplication. But I'm not sure if just a "completeness" argument warrants that change.

Aside from the above I also added implementations for most of the `RealFunction` except for `static func signGamma(_ x: Self) -> FloatingPointSign` since it can't implemented for simd types. So unfortunately we can't conform the simdtypes to the protocol. 

Let me know your thoughts/questions!

### DIFF
--- a/Sources/RealModule/SIMD+ElementaryFunctions.swift
+++ b/Sources/RealModule/SIMD+ElementaryFunctions.swift
@@ -1,16 +1,3 @@
-extension SIMD2: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
-extension SIMD4: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
-extension SIMD8: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
-extension SIMD16: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
-extension SIMD32: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
-extension SIMD64: @retroactive AdditiveArithmetic where Scalar: FloatingPoint {}
-
-extension SIMD2: ElementaryFunctions where Scalar: ElementaryFunctions & FloatingPoint { }
-extension SIMD4: ElementaryFunctions where Scalar: ElementaryFunctions & FloatingPoint { }
-extension SIMD8: ElementaryFunctions where Scalar: ElementaryFunctions & FloatingPoint { }
-extension SIMD16: ElementaryFunctions where Scalar: ElementaryFunctions & FloatingPoint { }
-extension SIMD32: ElementaryFunctions where Scalar: ElementaryFunctions & FloatingPoint { }
-extension SIMD64: ElementaryFunctions where Scalar: ElementaryFunctions & FloatingPoint { }
 
 extension SIMD where Scalar: ElementaryFunctions {
     @_transparent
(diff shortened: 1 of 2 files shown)


JaapWijnen:
@stephentyrone in case you hadn't seen this yet. Just wanted to see if you had any thoughts on this. No rush though!

stephentyrone:
I'm currently pondering this, and will likely have some questions about it after I finish tagging 1.1

JaapWijnen:
Of course! Let me know your thoughts once you have some time.

stephentyrone:
That might be the thing to do with this, yeah.

JaapWijnen:
@stephentyrone can we still consider merging these implementations? If you still have concerns after removing the `AdditiveArithmetic` conformances, please let me know

PR #321, opened 2025-08-11
Remove precondition assert from Augmented.sum(large:small:)

stephentyrone:
This is potentially useful to identify some bugs, but there are also algorithms (notably Kahan summation) that can be written in terms of fast2sum but would fail this check. Rework documentation to account for the change.


glessard (approved):
Note that on line 115, the world "summand" should be plural.

glessard:
@swift-ci please test

stephentyrone:
> Note that on line 115, the world "summand" should be plural.

Fixed in https://github.com/apple/swift-numerics/pull/265/commits/c3757dc2664da239db806114073a9e8a1c197ae2 (rolled into the documentation PR I have open already).

PR #330, opened 2025-09-18
Support Embedded Swift

ebariaux:
I'd like to be able to use Swift Numerics with code targeting the Embedded mode.
Excluding those 3 extensions allowed me to include the package and use ComplexModule in some code deployed to an STM32 board.

### DIFF
--- a/Sources/ComplexModule/Complex+Codable.swift
+++ b/Sources/ComplexModule/Complex+Codable.swift
@@ -12,6 +12,7 @@
 import RealModule
 
 // FloatingPoint does not refine Codable, so this is a conditional conformance.
+@_unavailableInEmbedded
 extension Complex: Decodable where RealType: Decodable {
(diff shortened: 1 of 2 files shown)


stephentyrone (approved):

stephentyrone:
@swift-ci test

Sources/ComplexModule/Complex+Codable.swift
```diff
@@ -12,6 +12,7 @@
 import RealModule
 
 // FloatingPoint does not refine Codable, so this is a conditional conformance.
+@_unavailableInEmbedded
```
stephentyrone, line 15: I believe that these will need to be wrapped as follows:

```suggestion
#if compiler(>=6.0)
@_unavailableInEmbedded
#endif
```

stephentyrone (changes requested):

